PR #135, opened 2023-06-26
fix(jdbc-postgres): void types from query are converted to null

### LINKED ISSUE #130: Postgres Void object are not serialized
```yaml
  - id: postgres
    type: io.kestra.plugin.jdbc.postgresql.Trigger
    interval: "PT5S"
    url: jdbc:postgresql://local:5432/kestra
    username: kestra
    password: k3str4
    fetch: true
    sql: "SELECT pg_sleep(15) as val"
```
lead to 
```
  - id: postgres
    type: io.kestra.plugin.jdbc.postgresql.Trigger
    interval: "PT5S"
    url: jdbc:postgresql://local:5432/kestra
    username: kestra
    password: k3str4
    fetch: true
    sql: "SELECT pg_sleep(15) as val"
```
should be simply null

brian-mulier-p:
closes #130


loicmathieu (approved):
Maybe other databases may take advantage of the `void` type.
Anyway, this will fix the issue we have today with Postgres.

PR #144, opened 2023-07-17
fix: allow setting the JDBC URL for MotherDuck

### LINKED ISSUE #143: Add MotherDuck support to the DuckDB plugin
### Feature description

MotherDuck should work out of the box based on the existing JDBC driver for DuckDB: https://motherduck.com/docs/integrations/jdbc-driver

image

But I can't connect, even when explicitly setting the token in the JDBC URL

```yaml
id: md
namespace: dev
tasks:
  - id: md
    sql: |
      SELECT passenger_count, avg(total_amount) as avg_fare
      FROM sample_data.nyc.yellow_cab_nyc_2022_11
      GROUP BY passenger_count
      ORDER by passenger_count;
    store: true

taskDefaults:
  - type: io.kestra.plugin.jdbc.duckdb.Query
    values:
      url: "jdbc:duckdb:md:my_db"
      timeZoneId: Europe/Berlin
```

A dedicated plugin with the MotherDuck logo and an easier way to pass the MOTHERDUCK_TOKEN from a secret would be quite useful 

The JDBC driver from here works well in my local IDE with both DuckDB and MotherDuck https://central.sonatype.com/artifact/org.duckdb/duckdb_jdbc/0.8.1/versions

loicmathieu:
Fixes #143

### DIFF
--- a/plugin-jdbc-duckdb/src/main/java/io/kestra/plugin/jdbc/duckdb/Query.java
+++ b/plugin-jdbc-duckdb/src/main/java/io/kestra/plugin/jdbc/duckdb/Query.java
@@ -101,7 +101,7 @@ public class Query extends AbstractJdbcQuery implements RunnableTask<Query.Outpu
     @Override
     @Schema(
         title = "The JDBC URL to connect to the database",
-        description = "The default value, `jdbc:duckdb:`, will use a local database. \nSet this property only when connecting to a MotherDuck instance, for example with `jdbc:duckdb:md:my_database?motherduck_token=<my_token>`.",
+        description = "The default value, `jdbc:duckdb:`, will use a local in-memory database. \nSet this property when connecting to a persisted database instance, for example `jdbc:duckdb:md:my_database?motherduck_token=<my_token>` to connect to [MotherDuck](https://motherduck.com/).",
         defaultValue = DEFAULT_URL
     )
     @NotNull
--- a/plugin-jdbc-duckdb/src/main/java/io/kestra/plugin/jdbc/duckdb/Query.java
+++ b/plugin-jdbc-duckdb/src/main/java/io/kestra/plugin/jdbc/duckdb/Query.java
@@ -24,6 +24,8 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
+import javax.validation.constraints.NotNull;
+
 import static io.kestra.core.utils.Rethrow.throwBiConsumer;
 
 @SuperBuilder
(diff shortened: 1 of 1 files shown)


plugin-jdbc-duckdb/src/main/java/io/kestra/plugin/jdbc/duckdb/Query.java
```diff
@@ -91,18 +99,30 @@ public void registerDriver() throws SQLException {
     }
 
     @Override
+    @Schema(
+        title = "The JDBC URL to connect to the database",
+        description = "The default value, `jdbc:duckdb:`, will use a local database. \nSet this property only when connecting to a MotherDuck instance, for example with `jdbc:duckdb:md:my_database?motherduck_token=<my_token>`.",
```
anna-geller, line 104:
```suggestion
        description = "The default value, `jdbc:duckdb:`, will use a local in-memory database. \nSet this property when connecting to a persisted database instance, for example `jdbc:duckdb:md:my_database?motherduck_token=<my_token>` to connect to [MotherDuck](https://motherduck.com/).",
```

anna-geller (approved):
thanks for the fix!

plugin-jdbc-duckdb/src/main/java/io/kestra/plugin/jdbc/duckdb/Query.java
```diff
@@ -91,18 +99,30 @@ public void registerDriver() throws SQLException {
...
     @Override
     public Query.Output run(RunContext runContext) throws Exception {
         Map<String, String> outputFiles = null;
 
-        this.databaseFile = runContext.tempFile();
-        Files.delete(this.databaseFile);
+        // we only create the database file if the default URL is used.
+        if (DEFAULT_URL.equals(this.url)) {
+            this.databaseFile = runContext.tempFile();
+            Files.delete(this.databaseFile);
```
brian-mulier-p, line 122: Why do we create a tempFile to delete it right afterwards (I know it was already there before but trying to understand)

brian-mulier-p (approved):
LGTM, just a little question

plugin-jdbc-duckdb/src/main/java/io/kestra/plugin/jdbc/duckdb/Query.java
loicmathieu: No idea, I think it's here to be sure that the file didn't exist as DuckDB will create it.

tchiotludo: duckdb expect to have an empty file for it's database, so this ugly method is only to have a temporary file path

PR #158, opened 2023-10-03
Update DuckDB to 0.9.0

armetiz:
I'm not a Java developper, sorry if I'm missing something.

Maven: https://central.sonatype.com/artifact/org.duckdb/duckdb_jdbc?smo=true
DuckDB Release: https://duckdb.org/2023/09/26/announcing-duckdb-090.html

I think that Kestra plugins does not need update.

Regards
thomas.

### DIFF
--- a/plugin-jdbc-duckdb/build.gradle
+++ b/plugin-jdbc-duckdb/build.gradle
@@ -13,7 +13,7 @@ jar {
 }
 
 dependencies {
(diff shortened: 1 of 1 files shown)


loicmathieu:
Unfortunatly we had to downgrade DuckDB to 0.8.1 due to an issue with the new driver version.

See https://github.com/kestra-io/plugin-jdbc/pull/164

PR #164, opened 2023-10-18
fix: downgrade DuckDB driver to 0.8.1

loicmathieu:
See https://github.com/duckdb/duckdb/issues/9383


loicmathieu:
I openned an issue to track at our side the upgrade to 0.9.0: https://github.com/kestra-io/plugin-jdbc/issues/165

PR #183, opened 2023-12-06
fix(pinot): correcting pinot trigger doc

shrutimantri:
The doc for Pinot trigger had an incorrect mention of waiting on Clickhouse DB, corrected it to be Pinot DB.


Skraye (approved):
LGTM

PR #260, opened 2024-03-22
Db2 plugin

iNikitaGricenko:

Implemented DB2 plugin (Which using db2 jdbc driver)
---

```yaml
id: read_db2
namespace: io.kestra.jdbc.db2

tasks:
  - id: read
    type: io.kestra.plugin.jdbc.db2.Query
    url: jdbc:db2://localhost:50000/testdb
    username: db2inst1
    password: password
    sql: select * from db2_types
    fetchOne: true
  - id: flow-id
    type: io.kestra.core.tasks.debugs.Return
    format: "{{outputs.update.row}}"

```
---

### Setup Instructions

- Setup docker from docker-compose-ci.yml file
- Run test


loicmathieu (approved):
LGTM, thanks

loicmathieu:
@iNikitaGricenko tests didn't pass, can yu have a look?

iNikitaGricenko:
Yes, seems like database doesn't have time to initialize before test runs

loicmathieu:
> Yes, seems like database doesn't have time to initialize before test runs

You can either configure a readiness check on the docker compose or add a little Thread.sleep(1000) so the test start later (ugly but quick)

iNikitaGricenko:
seems like it doesn't wait for healthcheck, so i'm adding thread.sleep to see if this help, to make sure that issue with database initialization (as it take too long for it to initialize)

PR #292, opened 2024-04-24
Sqlite

iNikitaGricenko:

Added code that will help sqlite to connect to database in kestra storage
---

Unfortunately i cannot test part where needs to download database file, and try to connect to it in external storage (Where file not in same directory as plugin). But hope that someone can help me with that

```yaml
id: sqlite-external
namespace: tutorial

tasks:
  - id: get0
    type: io.kestra.plugin.fs.http.Download
    disabled: true
    uri: https://github.com/lerocha/chinook-database/blob/master/ChinookDatabase/DataSources/Chinook_Sqlite.sqlite

  - id: wdir
    type: io.kestra.core.tasks.flows.WorkingDirectory
    tasks:
      - id: get
        type: io.kestra.plugin.scripts.shell.Commands
        outputFiles:
          - "*.sqlite"
        taskRunner:
          type: io.kestra.core.models.tasks.runners.types.ProcessTaskRunner
        commands:
          - curl -L -o Chinook_Sqlite.sqlite https://github.com/lerocha/chinook-database/raw/master/ChinookDatabase/DataSources/Chinook_Sqlite.sqlite
      - id: db
        type: io.kestra.core.tasks.storages.LocalFiles
        inputs:
          Chinook_Sqlite.sqlite: "{{ outputs.get.outputFiles['Chinook_Sqlite.sqlite'] }}"

      - id: analyze_media_store
        type: io.kestra.plugin.jdbc.sqlite.Query
        description: Find the total number of tracks purchased in each genre
        url: jdbc:sqlite:Chinook_Sqlite.sqlite
        sqliteFile: "{{ outputs.get.outputFiles['Chinook_Sqlite.sqlite'] }}"
        sql: |
          SELECT Genre.Name, COUNT(InvoiceLine.InvoiceLineId) AS TracksPurchased
          FROM Genre
          JOIN Track ON Genre.GenreId = Track.GenreId
          JOIN InvoiceLine ON Track.TrackId = InvoiceLine.TrackId
          GROUP BY Genre.Name
          ORDER BY TracksPurchased DESC;
```

---


plugin-jdbc-sqlite/src/main/java/io/kestra/plugin/jdbc/sqlite/Query.java
```diff
@@ -35,25 +42,79 @@
...
+                "  sqliteFile: {{ outputs.get.outputFiles['myfile.sqlite'] }}",
                 "  sql: select concert_id, available, a, b, c, d, play_time, library_record, floatn_test, double_test, real_test, numeric_test, date_type, time_type, timez_type, timestamp_type, timestampz_type, interval_type, pay_by_quarter, schedule, json_type, blob_type from pgsql_types",
                 "  fetch: true",
                 "- id: use-fetched-data",
                 "  type: io.kestra.plugin.jdbc.sqlite.Query",
                 "  url: jdbc:sqlite:myfile.db",
-                "  sql: \"{% for row in outputs.update.rows %} INSERT INTO pl_store_distribute (year_month,store_code, update_date) values ({{row.play_time}}, {{row.concert_id}}, TO_TIMESTAMP('{{row.timestamp_type}}', 'YYYY-MM-DDTHH:MI:SS.US') ); {% endfor %}\""}
+                "  sqliteFile: {{ outputs.get.outputFiles['myfile.sqlite'] }}",
+                "  sql: \"{% for row in outputs.update.rows %} INSERT INTO pl_store_distribute (year_month,store_code, update_date) values ({{row.play_time}}, {{row.concert_id}}, TO_TIMESTAMP('{{row.timestamp_type}}', 'YYYY-MM-DDTHH:MI:SS.US') ); {% endfor %}\""
+            }
```
loicmathieu, line 53: Can you provide a new example instead of updating the existing one?

plugin-jdbc-sqlite/src/main/java/io/kestra/plugin/jdbc/sqlite/Query.java
```diff
@@ -35,25 +42,79 @@
...
     @Override
     public Properties connectionProperties(RunContext runContext) throws Exception {
         Properties properties = super.connectionProperties(runContext);
 
+        URI url = URI.create((String) properties.get("jdbc.url"));
+
+        // get file name from url scheme parts
+        String filename = url.getSchemeSpecificPart().split(":")[1];
+
+        Path path = workingDirectory.toAbsolutePath().resolve(filename);
```
loicmathieu, line 79: There is a `runContext.resolve(path)` method that check for path traversal, please use it.

plugin-jdbc-sqlite/src/main/java/io/kestra/plugin/jdbc/sqlite/Query.java
```diff
@@ -35,25 +42,79 @@
...
+        if (path.toFile().canRead()) {
```
loicmathieu, line 80: You can remove this check, if we cannot read or write the file, we should throw

loicmathieu (changes requested):

plugin-jdbc-sqlite/src/main/java/io/kestra/plugin/jdbc/sqlite/Query.java
iNikitaGricenko: Can i replace it with ```path.toFile().exists()``` (as it more suitable in this case), so if it exists then we will use this file as database file, else create in memory database

loicmathieu: Yes

PR #301, opened 2024-05-08
Correct the url used in DuckDB query

shrutimantri:


wrussell1999 (commented):
Looks good but it’s currently failing the test. Seems unrelated to the change but will find out what’s causing it to fail before merging.

wrussell1999 (approved):

PR #302, opened 2024-05-08
fix(docs): change the arrow flight sql and IBM as400 logo

shrutimantri:
fix(docs): change the arrow flight sql and IBM as400 logo for dark background


wrussell1999:
@shrutimantri as the file names have changed, do these need to be referenced differently too?

wrussell1999:
Screenshot 2024-05-08 at 12 11 26
Undefined image because it doesn't exist

PR #306, opened 2024-05-16
fix(images): images suited for black background

shrutimantri:


wrussell1999 (changes requested):
```
You need to update the icon to handle black / white theme, example of wrong [one](https://raw.githubusercontent.com/kestra-io/kestra/6de50afeca9600dbf1c528f7daa5a3bf9a664e82/core/src/main/resources/icons/io.kestra.core.tasks.flows.WaitFor.svg), good [one](https://raw.githubusercontent.com/kestra-io/kestra/d28b63de0feef1ea629ce04e76987af4f6076c1d/core/src/main/resources/icons/io.kestra.core.tasks.flows.Pause.svg). The main difference is :
fill="currentColor"  
that allow to change the color of the svg.
Please take care of this?
```

tchiotludo:
Seems that was done in a meantime

PR #353, opened 2024-08-12
fix(example): correct the postgres copyin example

shrutimantri:


wrussell1999 (approved):

wrussell1999:
merging once CI has passed to confirm syntax is good

PR #390, opened 2024-10-07
Prevent resource leak by closing the PreparedStatement

### LINKED ISSUE #321: Task 'io.kestra.plugin.jdbc.oracle.Batch' fails with big number of rows (does not properly close cursors)
### Expected Behavior

I have a ION file that contains 1.6M rows (extracted via `io.kestra.plugin.jdbc.oracle.Query`).
I would like to insert these rows into a table, by using `io.kestra.plugin.jdbc.oracle.Batch`.

If I set the `chunk` to 1.000 (the default), I get an error `reactor.core.Exceptions$ReactiveException: java.sql.SQLException: ORA-01000: maximum open cursors exceeded`

If I set the `chunk` to 10.000 then the task finishes successfully.
This indicates that perhaps the Batch task does not properly close ResultSets.

I expect that the execution is successful for any batch size, and any number of rows.

Screenshot with successful execution (`chunk: 10000`):
image

Screenshot with error (`chunk: 1000`):
image

### Actual Behaviour

Execution fails with error if the default chunk size is used

### Steps To Reproduce

Use `io.kestra.plugin.jdbc.oracle.Batch` with 2M rows input

### Environment Information

- Kestra Version: 0.17.4
- Plugin version: default
- Operating System (OS / Docker / Kubernetes): Kubernetes
- Java Version (If not docker): default

### Example flow

My yaml looks like this (I've omitted the inputs and the previous tasks which compute the column names):

```
  # Select all the rows from the source table
  - id: select
    type: io.kestra.plugin.jdbc.oracle.Query
    store: true
    sql: |
      SELECT 
        {% for col in outputs.metadata.rows -%}
          {{ col.COLUMN_NAME }}
          {% if not loop.last -%},{% endif -%}
        {% endfor -%}
      FROM {{ inputs.srcSchema}}.{{inputs.srcTable}}

  # Insert all the rows into the target table
  - id: insert
    type: io.kestra.plugin.jdbc.oracle.Batch      
    from: "{{outputs.select.uri}}"
    chunk: 1000
    sql: |
      INSERT INTO {{inputs.destTable}} (
        {% for col in outputs.metadata.rows -%} 
          {{ col.COLUMN_NAME }} {% if not loop.last %},{% endif %}
        {% endfor -%}
      ) VALUES (
        {% for col in outputs.metadata.rows %} ? {% if not loop.last %},{% endif %} {% endfor %}
      )

```

lburja:
fix(jdbc): avoid resource leak when using batching, by closing the PreparedStatement

close  #321

### DIFF
--- a/plugin-jdbc/src/main/java/io/kestra/plugin/jdbc/AbstractJdbcBatch.java
+++ b/plugin-jdbc/src/main/java/io/kestra/plugin/jdbc/AbstractJdbcBatch.java
@@ -117,20 +117,21 @@ public abstract class AbstractJdbcBatch extends Task implements JdbcStatementInt
                 })
                 .buffer(this.chunk, this.chunk)
                 .map(throwFunction(o -> {
-                    PreparedStatement ps = connection.prepareStatement(sql);
-                    ParameterType parameterMetaData = ParameterType.of(ps.getParameterMetaData());
+                    try (PreparedStatement ps = connection.prepareStatement(sql)) {
+                        ParameterType parameterMetaData = ParameterType.of(ps.getParameterMetaData());
 
-                    for (Object value : o) {
-                        ps = this.addRows(ps, parameterMetaData, value, cellConverter, connection);
+                        for (Object value : o) {
+                            this.addRows(ps, parameterMetaData, value, cellConverter, connection);
 
-                        ps.addBatch();
-                        ps.clearParameters();
+                            ps.addBatch();
+                            ps.clearParameters();
+                        }
+
+                        int[] updatedRows = ps.executeBatch();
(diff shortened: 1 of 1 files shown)


loicmathieu (approved):
Thanks, this looks good to me.

Did you tested it with your file that contains 1.6M rows ?

lburja:
Yes I did. I modified the test `BatchTest::namedInsert` to insert 2.000.000 rows into a table instead of 5.

As expected, the test fails:
image

Then, I applied the fix, and now the test passes:
image

loicmathieu:
@lburja great! Thanks for the confirmation

Issue #368, opened 2024-08-28
Add a new Queries task to run multiple SQL with multiple outputs, parameter binding and transactions


anna-geller:
### **Context**

The current `Query` task is limited to executing a single SQL statement and handling its output. Some automation tasks require executing multiple SQL statements, potentially wrapped in a transaction, while handling outputs of multiple `SELECT` statements. 

To avoid breaking changes in the `Query` tasks, here is a proposal for a new `Queries` task.

This task will enable:
- execution of multiple SQL statements
- managing outputs of multiple SELECT queries
- handling parameters to prevent SQL injection
- transactions.

### Use case for testing

Start Postgres in a container:

```sql
docker run --name queries -e POSTGRES_PASSWORD=mysecretpassword -e POSTGRES_USER=postgres -e POSTGRES_DB=postgres -p 5432:5432 -d postgres
```

Create tables (alternatively directly from the new task):

```sql
CREATE TABLE myusers (
    id SERIAL PRIMARY KEY,
    name VARCHAR(255) NOT NULL,
    email VARCHAR(255) UNIQUE NOT NULL,
    last_login TIMESTAMP
);

CREATE TABLE mylogs (
    log_id SERIAL PRIMARY KEY,
    user_email VARCHAR(255) NOT NULL,
    action VARCHAR(255) NOT NULL,
    timestamp TIMESTAMP NOT NULL,
    FOREIGN KEY (user_email) REFERENCES myusers(email)
);
```

### **Implementation**

**Task Type**: `io.kestra.plugin.jdbc.postgresql.Queries` (equivalent tasks will be needed for all JDBC plugin subgroups, not just Postgres)

**Description**: The `Queries` task allows executing multiple SQL statements within a single task, with support for parameter binding and transaction management. The task can handle multiple `SELECT` statements and their outputs, allowing you to fetch the results directly or store them as internal storage ION files.

### **Properties**

- **sql**:
    - **Type**: `string`
    - **Required**: ✔️
    - **Description**: A single string containing multiple SQL statements separated by semicolons `;`. The statements will be executed sequentially (optionally as part of a transaction).
- **parameters**:
    - **Type**: `map`
    - **Required**: ❌
    - **Description**: A map of parameters to bind to the SQL queries. The keys should match the parameter placeholders in the SQL string, e.g., `:parameterName`.
- **transaction**:
    - **Type**: `boolean`
    - **Required**: ❌
    - **Default**: `false`
[...]

PR #398, opened 2024-10-14
feat: Add a new Queries task to run multiple SQL with multiple outputs, parameter binding and transactions

mgabelle:
extract common logic to utils and abstract classes

created test for queries

implemented queries for MySQL

#368


build.gradle
```diff
@@ -67,6 +67,9 @@ subprojects {
 
         // Logs
         compileOnly'org.slf4j:slf4j-api'
+
+        //SQL named parameters binding
+        implementation group: 'org.httprpc', name: 'httprpc', version: '5.9'
```
loicmathieu, line 72: This seems to be a full fledge HTTP server and client framework, this is a lot to import just to have named parameter injection.
Moreover, it seems to be a one-man project so I'm not fan to include it.
It also seems to moved to https://github.com/HTTP-RPC/Kilo

PR #425, opened 2024-10-31
refactor(queries): change parser to use JSQLParser

mgabelle:


AcevedoR:
hey is it still pertinent ? what does it mean to fix/improve ?

mgabelle:
Indeed it's not relevant anymore.
At this point I was doing some testing for the Queries tasks. 
I will close the PR. 👍🏻

PR #535, opened 2025-02-27
Fix/gh 528

fhussonnois:


fhussonnois:
@loicmathieu Do you know if there's a specific reason why we created a PreparedStatement per batch instead of just creating it once?

PR #539, opened 2025-03-03
Support hstore data type in postgres

### LINKED ISSUE #538: Support hstore data type in Postgres
### Feature description

`hstore` is a PostgreSQL extension that allows storing key-value pairs in a single column. There is currently not support for this, with an error being thrown when attempting to query a table with a `hstore` data type.

With the data type being a list of key value pairs it would potentially be good to convert this to a HashMap but most ETL tools like Airbyte treat it as a string which is probably a better type to go with, users can always process downstream if needed.

I'll work on a PR for this

benpops89:
Support the [`hstore`](https://www.postgresql.org/docs/current/hstore.html) data type in Postgres, returning the value as a string, closes #538

---

I tested the changes by spinning up a local postgres instance and creating a table with a `hstore` data type and then running a simple query flow 

```sql
CREATE EXTENSION IF NOT EXISTS hstore;

CREATE TABLE products (
    id SERIAL PRIMARY KEY,
    name TEXT NOT NULL,
    attributes HSTORE
);

INSERT INTO products (name, attributes) VALUES ('Laptop', 'processor=>"Intel i7", RAM=>"16GB"');
```

```yaml
id: hstore
namespace: test

tasks:
  - id: query
    type: io.kestra.plugin.jdbc.postgresql.Query
    url: jdbc:postgresql://172.17.0.2:5432/demo
    username: postgres
    password: password
    sql: SELECT * FROM products
    fetchType: FETCH
```

The results from the query are as follows
```json
{
    "id": 1,
    "name": "Laptop",
    "attributes": "\"RAM\"=>\"16GB\", \"processor\"=>\"Intel i7\""
}
```

### DIFF
--- a/plugin-jdbc-postgres/src/main/java/io/kestra/plugin/jdbc/postgresql/PostgresCellConverter.java
+++ b/plugin-jdbc-postgres/src/main/java/io/kestra/plugin/jdbc/postgresql/PostgresCellConverter.java
@@ -48,6 +48,8 @@ public class PostgresCellConverter extends AbstractCellConverter {
                 PGInterval interval = (PGInterval) data;
                 // Returns an iso 8601 duration format
                 return getISO8601Interval(interval.getYears(), interval.getMonths(), interval.getDays(), interval.getHours(), interval.getMinutes(), (int) interval.getSeconds());
+            case "hstore":
+                return rs.getString(columnIndex);
         }
 
         Class<?> clazz = data.getClass();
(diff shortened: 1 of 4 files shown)


loicmathieu (changes requested):
Hi, thanks a lot for your contribution.

Do you think it would be possible to add a unit test?

I'm not familiar with hstore, but if it's designed to be able to store large value maybe reading it as a string is not the most convenient way. For example CLOB and BLOB are read as output streams. Anyway, if you think String is enough I'm OK with it.

benpops89:
@loicmathieu - thanks for the feedback.

`hstore` is not designed to hold large values, it's optimised for storing key-value pairs of relatively small text-based values. If you needed to store larger values `JSONB` would be a better type.

Thinking about this some more returning a `Map<String, String>` would make more sense than a `String`. I'm going to update and I'll look at adding some tests as well

benpops89:
@loicmathieu - I've added the tests for hstore and they are passing locally.

loicmathieu (approved):
LGTM, thanks

PR #557, opened 2025-03-18
chore: try to fix CI by bumping slf4j

AcevedoR:
try


AcevedoR:
already fixed in #559

PR #561, opened 2025-03-20
feat(duckdb): workingDir syntax not needed anymore for input files

### LINKED ISSUE #267: Find a way to not require {{ workingDir }} within a DuckDB query when input files are provided
### Feature description

See an example:

```yaml
id: api_python_sql
namespace: blueprint

tasks:
  - id: extract
    type: io.kestra.plugin.fs.http.Download
    uri: https://dummyjson.com/products

  - id: transform
    type: io.kestra.plugin.scripts.python.Script
    docker:
      image: python:3.11-alpine # 57 MB
    inputFiles:
      data.json: "{{ outputs.extract.uri }}"
    outputFiles:
      - "*.json"
    script: |
      import json

      with open("data.json", 'r') as file:
        data = json.load(file)

      filtered_data = [{"brand": product["brand"], "price": product["price"]} for product in data["products"]]
      with open("products.json", 'w') as file:
        json.dump(filtered_data, file, indent=4)

  - id: sqlQuery
    inputFiles:
      products.json: "{{ outputs.transform.outputFiles['products.json'] }}"
    sql: |
      INSTALL json;
      LOAD json;
      SELECT brand, round(avg(price), 2) as avg_price
      FROM read_json_auto('{{ workingDir }}/products.json') # ❌ this is error-prone and not intuitive
      GROUP BY brand
      ORDER BY avg_price DESC;
    store: true
```

Ideally, the same should work by using:

```yaml
...
  - id: sqlQuery
    inputFiles:
      products.json: "{{ outputs.transform.outputFiles['products.json'] }}"
    sql: |
      INSTALL json;
      LOAD json;
      SELECT brand, round(avg(price), 2) as avg_price
      FROM read_json_auto('products.json')
      GROUP BY brand
      ORDER BY avg_price DESC;
    store: true
```

AcevedoR:
unfortunately could not use [JDBC connection properties](https://duckdb.org/docs/stable/clients/java#configuring-connections) because `file_search_path` [is not a DuckDB global option](https://duckdb.org/docs/stable/configuration/overview.html#local-configuration-options)

- fixes #267

### DIFF
--- a/plugin-jdbc-duckdb/src/main/java/io/kestra/plugin/jdbc/duckdb/DuckDbQueryInterface.java
+++ b/plugin-jdbc-duckdb/src/main/java/io/kestra/plugin/jdbc/duckdb/DuckDbQueryInterface.java
@@ -10,7 +10,7 @@ public interface DuckDbQueryInterface {
     @Schema(
         title = "Input files to be loaded from DuckDb.",
         description = "Describe a files map that will be written and usable by DuckDb. " +
-            "You can reach files using a `workingDir` variable, example: `SELECT * FROM read_csv_auto('{{ workingDir }}/myfile.csv');` "
+            "You can reach files by their filename, example: `SELECT * FROM read_csv_auto('myfile.csv');` "
     )
     @PluginProperty(
         additionalProperties = String.class,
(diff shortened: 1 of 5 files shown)


loicmathieu (approved):
LGTM, one small note, instead of crafting a multi-SQL query, would it be possible to execute two queries instead or would it be too difficult

I'm OK with the current solution so if it's not easy, keep it like that.

AcevedoR:
> LGTM, one small note, instead of crafting a multi-SQL query, would it be possible to execute two queries instead or would it be too difficult
> 
> I'm OK with the current solution so if it's not easy, keep it like that.

I tried looking at Queries to see if I could reuse some of that:
https://github.com/kestra-io/plugin-jdbc/blob/af93464134dd576150d11a92a13aa5dbfe86aede/plugin-jdbc/src/main/java/io/kestra/plugin/jdbc/AbstractJdbcQueries.java#L63-L81

but since it was already very simple and not reusable it seems keeping things simple was maybe cleaner

PR #566, opened 2025-03-21
chore: depreciate autocommit property

### LINKED ISSUE #550: Deprecate autocommit property
### Describe the issue

We want to deprecate `autocommit` on both Query and Queries tasks because:
1. `Query` task is a single query, not a transaction so autocommit is no longer needed
2. `Queries` task automatically processes all queries as one transaction so autocommit property is ignored anyway (everything processed within Queries is treated as a transaction)

### Environment

- Kestra Version: develop

AcevedoR:
first part of #550
only doing the documentation depreciation, will remove conditional logic with Store later so it can be more tested/QA


AcevedoR:
- having a weird issue, the deprecation is not showing, maybe linked to #567

AcevedoR:
> * having a weird issue, the deprecation is not showing, maybe linked to [PostsgreSQL Query.fetchSize documentation is missing #567](https://github.com/kestra-io/plugin-jdbc/issues/567)

this was fixed, deprecation is showing now

PR #578, opened 2025-03-31
feat(snowflake): allow using encrypted private key for auth

AcevedoR:


nkwiatkowski:
Looks good, but there are no unit test on this new feature. Maybe adding a test to validate both encryption and decryption should be useful

AcevedoR:
> Looks good, but there are no unit test on this new feature. Maybe adding a test to validate both encryption and decryption should be useful

added tests @nkwiatkowski wdyt ?

plugin-jdbc-snowflake/build.gradle
```diff
@@ -15,6 +15,8 @@ jar {
 dependencies {
     implementation("net.snowflake:snowflake-jdbc:3.23.1")
     implementation project(':plugin-jdbc')
+    implementation("org.bouncycastle:bcprov-jdk18on:1.80")
```
loicmathieu, line 18: It is already managed in the platform

plugin-jdbc-snowflake/build.gradle
```diff
@@ -15,6 +15,8 @@ jar {
...
+    implementation("org.bouncycastle:bcpkix-jdk18on:1.80")
```
loicmathieu, line 19: This one is not on the platform, please add it their and remove the version from here.

plugin-jdbc-snowflake/src/main/java/io/kestra/plugin/jdbc/snowflake/RSAKeyPairUtils.java
```diff
@@ -0,0 +1,50 @@
...
+import java.security.spec.InvalidKeySpecException;
+import java.security.spec.PKCS8EncodedKeySpec;
+import java.util.Base64;
+import java.util.Optional;
+
+public class RSAKeyPairUtils {
+    public static PrivateKey deserializePrivateKey(String privateKey, Optional<String> privateKeyPassword) {
+        Security.addProvider(new BouncyCastleProvider());
+        var keyBytes = Base64.getDecoder().decode(privateKey);
+        try {
```
loicmathieu, line 22: I'm not fan of try/catch as a "normal path", better is to either add a property to say wether or not it's encrypted, or base on the fact that a password exist so it must be encrypted.

loicmathieu (changes requested):

PR #635, opened 2025-06-24
fix(mysql): sql is not required

### LINKED ISSUE #523: Batch query sql argument are not required
on this [page](https://kestra.io/plugins/plugin-jdbc-mysql/io.kestra.plugin.jdbc.mysql.batch), sql is required 
image

But the description of `table` argument claims it's not required: 
> The table from which column names will be retrieved.
> 
> This property specifies the table name which will be used to retrieve the columns for the inserted values. You can use it instead of specifying manually the columns in the columns property. In this case, the sql property can also be omitted, an INSERT statement would be generated automatically.

Validate each case and adapt the documentation generator to be relevant

fdelbrayelle:
closes #523 

---

---


fdelbrayelle:
@Malaydewangan09  Yes you're right I'm adding it back! Thanks!

PR #897, opened 2026-05-11
fix(tests): harden ClickHouse timeout and Druid ingestion retry in CI

fdelbrayelle:
## Summary

- **ClickHouse `RunnerTest`**: the default `@ExecuteFlow` timeout (~5 min) was racing with the `wait_for_clickhouse` retry loop (`maxAttempt: 60 × interval: PT5S = 5 min`) on slow CI runners. Raised the timeout to `PT600S` (10 min), matching the pattern already used by `db2` and `sybase`.
- **Druid `DruidTestHelper`**: `runInlineIngestion()` was throwing immediately on a `500` response wrapping a `503 Service Unavailable` from the Overlord. The Overlord is still warming up even after the router and coordinator `/status` checks return 200. Wrapped the ingestion submission in an `Await.until` retry loop (5 s interval, 5 min total) so transient startup 503s are absorbed instead of failing the test setup.

## Test plan

- [ ] CI passes on this branch with no ClickHouse or Druid test failures
- [ ] Druid `DruidTest` and `DruidTriggerTest` pass consistently
- [ ] ClickHouse `RunnerTest.all_clickhouse` passes consistently

### DIFF
--- a/plugin-jdbc-clickhouse/src/test/java/io/kestra/plugin/jdbc/clickhouse/RunnerTest.java
+++ b/plugin-jdbc-clickhouse/src/test/java/io/kestra/plugin/jdbc/clickhouse/RunnerTest.java
@@ -14,7 +14,7 @@ import static org.hamcrest.Matchers.is;
 class RunnerTest {
 
     @Test
-    @ExecuteFlow("sanity-checks/all_clickhouse.yaml")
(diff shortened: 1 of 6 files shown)


Malaydewangan09:
As discussed [here](https://kestra-io.slack.com/archives/C0B1RQWV09G/p1778503261897009?thread_ts=1778503197.085129&cid=C0B1RQWV09G), we will disable the Druid tests.

Malaydewangan09 (approved):

PR #921, opened 2026-06-08
fix(core): migrate FileSerde to InputStream/OutputStream API for binary ION support

Malaydewangan09:
Migrates the FileSerde reads and writes to byte streams so binary ION is not corrupted.


fdelbrayelle (approved):
LGTM but CI is failing

Malaydewangan09:
Yes, failing on trivy step not tests

PR #924, opened 2026-06-15
perf(jdbc): connection pooling + hoist ResultSet metadata (#909)

### LINKED ISSUE #909: perf: no connection pooling (DriverManager per task) and redundant getMetaData() calls per row
## Summary

This issue tracks two performance problems found during an audit of the plugin. They are grouped here to avoid noise; each can be addressed independently.

---

### 1. No connection pool — `DriverManager.getConnection()` per task execution
**File**: `plugin-jdbc/src/main/java/io/kestra/plugin/jdbc/JdbcConnectionInterface.java:102`

Every task run calls `DriverManager.getConnection(jdbcUrl, props)`, which opens a fresh TCP connection, performs a full TLS handshake, and tears it down at end of scope. For short query tasks run at high frequency this adds 50–200ms of pure connection overhead per execution.

**Estimated gain**: 40–80% wall-clock reduction for short-running queries.

**Fix**: Integrate HikariCP (`HikariDataSource`) keyed by the rendered JDBC URL, cached in a `ConcurrentHashMap<String, HikariDataSource>` within a `JdbcConnectionPool` helper. `connection()` acquires from the pool rather than calling `DriverManager` directly. Pool size should be configurable with a default of 5–10.

---

### 2. `rs.getMetaData()` called N×C times per result set
**File**: `plugin-jdbc/src/main/java/io/kestra/plugin/jdbc/AbstractJdbcBaseQuery.java:187,191`

In `mapResultSetToMap()`, `rs.getMetaData()` is called once for `getColumnCount()` and again inside the column loop for every `getColumnLabel(i)` call. For a result set with 100 columns and 100,000 rows that is 10 million redundant metadata retrievals. On many JDBC drivers `getMetaData()` is not free.

**Estimated gain**: 20–40% CPU reduction on wide tables with large result sets.

**Fix**: Hoist `ResultSetMetaData meta = rs.getMetaData()` before the row loop. Extract column labels into a `String[] labels` array before the first `rs.next()` call since labels are invariant across rows.

Malaydewangan09:
- closes #909


Malaydewangan09:
# QA: connection pooling + ResultSet metadata hoist (#909)

Two independent performance fixes from the issue audit. Each is measured with JMH against a real driver.

## Part 2 - what changed (metadata hoist)

`mapResultSetToMap` read the result-set metadata on every row, once for the column count and again for every column label:

```java
// before (runs per row)
int n = rs.getMetaData().getColumnCount();
for (int i = 1; i <= n; i++) {
    map.put(rs.getMetaData().getColumnLabel(i), convertCell(i, rs, ...));
}

// after: read metadata once per ResultSet, reuse the labels for every row
String[] labels = columnLabels(rs);            // getMetaData() called once
for (int i = 1; i <= labels.length; i++) {
    map.put(labels[i - 1], convertCell(i, rs, ...));
}
```

Column labels do not change between rows, so they are computed once per result set instead of once per row. The old cost scales as O(rows x columns), the new cost as O(columns) per result set.

### Benchmark (JMH)

JMH 1.37, JDK 25 toolchain, Throughput mode, 3 forks x 8 iterations, real H2 driver, 1000 rows. Higher is better.

| Columns | Before (ops/ms) | After (ops/ms) | Change |
|---|---:|---:|---:|
| 10  | 5.35 +/- 0.20 | 9.34 +/- 0.37 | +75%, ~1.75x |
| 50  | 0.96 +/- 0.07 | 1.30 +/- 0.13 | +35%, ~1.35x |
| 100 | 0.47 +/- 0.00 | 0.59 +/- 0.05 | +27%, ~1.27x |

Allocation per row is identical between before and after (`gc.alloc.rate.norm` from the gc profiler). This is a pure CPU win, not a memory one. The gain grows with how expensive a driver's `getMetaData()` is, so H2 is a conservative case.

## Part 1 - what changed (connection pooling)

Every task opened a fresh connection and tore it down at the end of the run:

```java
// before: new TCP connection (plus TLS handshake on remote) every task run
return DriverManager.getConnection(jdbcUrl, props);

// after: reuse a pooled connection keyed by URL + credentials
return JdbcConnectionPool.connection(jdbcUrl, props, size);
```

Pooling is on by default for remote drivers, excluded for embedded drivers (DuckDB, SQLite, MS Access), and can be turned off per task with `connectionPooling: false`. Pool size is configurable via `connectionPoolSize` (default 10).

### Benchmark (JMH)

JMH 1.37, JDK 25 toolchain, AverageTime and SampleTime modes, microseconds, 2 forks x 5 iterations, `-prof gc`. Real PostgreSQL 16 in a Testcontainers container. One operation is: acquire a connection, run `SELECT 1`, close it. Lower is better.

| Scenario | Metric | Before (DriverManager) | After (HikariCP) | Change |
|---|---|---:|---:|---:|
| 1 thread   | Time per op   | 2270 +/- 167 us  | 116 +/- 13 us  | 95% lower, ~20x faster |
| 1 thread   | Memory per op | 280,703 B        | 1,291 B        | 99.5% lower, ~217x less garbage |
| 16 threads | Time per op   | 16,216 +/- 4224 us | 660 +/- 143 us | 96% lower, ~25x faster |
| 16 threads | Memory per op | 281,628 B        | 1,472 B        | 99.5% lower, ~191x less garbage |

--- 

Screenshot 2026-06-16 at 2 03 26 PM


The Postgres container runs on localhost with no TLS, so the saving measured here is only the local TCP and auth handshake. A real remote database with TLS pays far more per connection, so these numbers are a conservative floor, not the headline. Under 16 threads against a pool of 10, six requests wait for a free connection, which lifts the pooled p99 tail. Sizing `connectionPoolSize` to the expected concurrency removes that tail.

## Environment

| Item | Value |
|---|---|
| CPU | Apple M4, 10 cores |
| OS | macOS 15.5 (Darwin 24.5.0), aarch64 |
| Benchmark JVM | OpenJDK 25.0.1 (Temurin) |
| JMH | 1.37 |
| H2 | 2.3.232 (in-memory) |
| PostgreSQL | postgres:16-alpine via Testcontainers 1.20.4 |
| HikariCP | 6.2.1 |

fdelbrayelle (approved):

PR #926, opened 2026-06-17
fix(jdbc-sqlite): handle ISO-8601 T-separated DATETIME values with fractional second

### LINKED ISSUE #15962: SQLite bas parsing DATETIME with milllisecond
### Describe the issue

Hi, with "io.kestra.plugin.jdbc.sqlite.Query", i need to extract data with a simple "SELECT *" to insert result in a Postgres db. With DATETIME columns, the regex used not include milliseconds format and it raise an error:
Image
Could you check this or add a property to change this?

### Environment

- Kestra Version: 1.3.16

suraj-0628:
Fixes https://github.com/kestra-io/kestra/issues/15962

## Problem
`io.kestra.plugin.jdbc.sqlite.Query` throws `Error parsing time stamp` when a
`DATETIME`/`TIMESTAMP` column stores an ISO-8601 value with a `T` separator,
e.g. `2012-12-26T12:06:28.727`.

## Root cause
`SqliteCellConverter` delegated all datetime parsing to
`ResultSet#getTimestamp(int, Calendar)`. The sqlite-jdbc driver parses
SQLite TEXT storage using `SimpleDateFormat` with the fixed pattern
`"yyyy-MM-dd HH:mm:ss.SSS"`, which contains a literal space and cannot
match an ISO-8601 `T` separator regardless of fractional seconds precision.

## Fix
When the underlying SQLite storage class is TEXT (`data instanceof String`),
parse the value directly in `SqliteCellConverter` using a `DateTimeFormatter`
that handles both `T` and space separators with 0–9 fractional digits, and
applies the task's configured `zoneId` explicitly.

INTEGER (unix epoch) and REAL (Julian day) storage are unaffected and
continue to use the driver's `getTimestamp()` path.

## Test assertion correction
The existing `select()` and `selectMustUseConfiguredTimezoneEvenIfJvmTimezoneIsUtc()`
assertions for `datetime_column` were updated from `2023-10-30T22:02:27.150Z`
to `2023-10-30T21:59:57.150150Z`.

The old value was produced by `SimpleDateFormat` in lenient mode reading
the 6-digit fraction `150150` as 150,150 milliseconds (2 min 30 sec),
rolling the time forward incorrectly. The new value is correct:
`2023-10-30 22:59:57 Europe/Paris (CET = UTC+1) = 21:59:57 UTC`,
with full microsecond precision now preserved.

## Note
`selectFromExistingDatabase`, `selectFromExistingDatabaseAndOutputDatabase`,
and `outputDbFileTrueWithoutSqliteFile_shouldCreateAndOutputDatabase` fail
on Windows due to a pre-existing `URISyntaxException` in `SqliteQueryUtils`
(backslash in JDBC URL). This is unrelated to this change.

### DIFF
--- a/plugin-jdbc-sqlite/src/main/java/io/kestra/plugin/jdbc/sqlite/SqliteCellConverter.java
+++ b/plugin-jdbc-sqlite/src/main/java/io/kestra/plugin/jdbc/sqlite/SqliteCellConverter.java
@@ -11,7 +11,27 @@ import java.time.ZoneId;
 import java.util.Calendar;
 import java.util.TimeZone;
 
+import java.time.Instant;
+import java.time.LocalDateTime;
+import java.time.ZonedDateTime;
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeFormatterBuilder;
+import java.time.temporal.TemporalAccessor;
+
 public class SqliteCellConverter extends AbstractCellConverter {
+
+    private static final DateTimeFormatter SQLITE_DATETIME_FORMATTER =
+    new DateTimeFormatterBuilder()
+        .appendOptional(DateTimeFormatter.ISO_ZONED_DATE_TIME)
+        .appendOptional(DateTimeFormatter.ISO_OFFSET_DATE_TIME)
+        .appendOptional(DateTimeFormatter.ISO_LOCAL_DATE_TIME)
+        .appendOptional(new DateTimeFormatterBuilder()
+            .append(DateTimeFormatter.ISO_LOCAL_DATE)
+            .appendLiteral(' ')
+            .append(DateTimeFormatter.ISO_LOCAL_TIME)
+            .toFormatter())
+        .appendOptional(DateTimeFormatter.ISO_LOCAL_DATE)
+        .toFormatter();
 
     public SqliteCellConverter(ZoneId zoneId) {
         super(zoneId);
@@ -29,7 +49,15 @@ public class SqliteCellConverter extends AbstractCellConverter {
 
         return switch (columnTypeName.toLowerCase()) {
             case "date" -> LocalDate.parse(resultSet.getString(columnIndex));
+            // case "datetime", "timestamp" -> {
+            //     var calendar = Calendar.getInstance(TimeZone.getTimeZone(zoneId));
+            //     var timestamp = resultSet.getTimestamp(columnIndex, calendar);
+            //     yield timestamp != null ? timestamp.toInstant() : null;
+            // }
             case "datetime", "timestamp" -> {
+                if (data instanceof String text) {
+                yield parseTextToInstant(text);
+                }
                 var calendar = Calendar.getInstance(TimeZone.getTimeZone(zoneId));
                 var timestamp = resultSet.getTimestamp(columnIndex, calendar);
                 yield timestamp != null ? timestamp.toInstant() : null;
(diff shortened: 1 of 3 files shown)


plugin-jdbc-sqlite/src/main/java/io/kestra/plugin/jdbc/sqlite/SqliteCellConverter.java
```diff
@@ -29,7 +49,15 @@ public Object convertCell(int columnIndex, ResultSet resultSet, Connection conne
 
         return switch (columnTypeName.toLowerCase()) {
             case "date" -> LocalDate.parse(resultSet.getString(columnIndex));
+            // case "datetime", "timestamp" -> {
```
fdelbrayelle: Remove commented block

plugin-jdbc-sqlite/src/test/java/io/kestra/plugin/jdbc/sqlite/SqliteTest.java
```diff
@@ -90,13 +91,42 @@ void selectMustUseConfiguredTimezoneEvenIfJvmTimezoneIsUtc() throws Exception {
                 .build();
 
             var runOutput = task.run(runContext);
-            assertThat(runOutput.getRow().get("datetime_column"), is(Instant.parse("2023-10-30T22:02:27.150Z")));
+            // assertThat(runOutput.getRow().get("datetime_column"), is(Instant.parse("2023-10-30T22:02:27.150Z")));
```
fdelbrayelle: Remove commented line

plugin-jdbc-sqlite/src/test/java/io/kestra/plugin/jdbc/sqlite/SqliteTest.java
```diff
@@ -63,7 +63,8 @@ void select() throws Exception {
         assertThat(runOutput.getRow().get("int_column"), is(42));
 
         assertThat(runOutput.getRow().get("date_column"), is(LocalDate.parse("2023-10-30")));
-        assertThat(runOutput.getRow().get("datetime_column"), is(Instant.parse("2023-10-30T22:02:27.150Z")));
+        // assertThat(runOutput.getRow().get("datetime_column"), is(Instant.parse("2023-10-30T22:02:27.150Z")));
```
fdelbrayelle: Remove commented line

plugin-jdbc-sqlite/src/test/resources/scripts/sqlite.sql
```diff
@@ -47,4 +47,30 @@ INSERT INTO lite_types (
     '{"key": "value"}',
     X'0102030405060708',
     NULL
+);
```
fdelbrayelle, line 50: This file doesn't seem to be used, if so please delete it (and its folder if empty then)

plugin-jdbc-sqlite/src/main/java/io/kestra/plugin/jdbc/sqlite/SqliteCellConverter.java
```diff
@@ -29,7 +49,15 @@ public Object convertCell(int columnIndex, ResultSet resultSet, Connection conne
...
+            //     var calendar = Calendar.getInstance(TimeZone.getTimeZone(zoneId));
+            //     var timestamp = resultSet.getTimestamp(columnIndex, calendar);
+            //     yield timestamp != null ? timestamp.toInstant() : null;
+            // }
             case "datetime", "timestamp" -> {
+                if (data instanceof String text) {
+                yield parseTextToInstant(text);
```
fdelbrayelle: missing tab for formatting

plugin-jdbc-sqlite/src/test/resources/scripts/sqlite.sql
suraj-0628: sqlite.sql is actively used by SqliteTest.java (line 435) and SqliteTriggerTest.java (line 45) via executeSqlScript("scripts/sqlite.sql") in their initDatabase() methods.

fdelbrayelle (approved):

PR #930, opened 2026-06-17
fix(trigger): default JDBC triggers to FETCH so they fire out of the box

jymaire:
## Summary

JDBC triggers (`io.kestra.plugin.jdbc.*.Trigger`) inherited `fetchType = NONE` from `AbstractJdbcTrigger`. `AbstractJdbcQuery#run` has no `NONE` arm in its `switch`, so no rows were read and `size` stayed `0`. `AbstractJdbcTrigger#evaluate` returns `Optional.empty()` whenever `size == 0`, so a trigger configured without an explicit `fetchType` never fired — the scheduler reported it as blocked ("No execution found, schedule is blocked since...").

- Override the `fetchType` default in `AbstractJdbcTrigger` from `NONE` to `FETCH`, so a trigger with only the required fields (`url`, `username`, `password`, `sql`) works out of the box and exposes `{{ trigger.rows }}`.
- Update the field `@Schema` to document the `FETCH` default and that `NONE` would never fire.
- Backward compatible: explicit `fetchType` values and the deprecated `fetch`/`fetchOne`/`store` booleans are unaffected — `renderFetchType()` short-circuits on those booleans before reading `fetchType`.

fixes: https://github.com/kestra-io/kestra-ee/issues/7414

## Test plan

- [x] New TDD test `PgsqlTriggerDefaultFetchTypeTest#firesWithoutExplicitFetchType` — a postgres trigger with no `fetchType` set must produce an execution. Fails before the fix (1-min poll latch times out, no execution), passes after.
- [x] `rtk test ./gradlew :plugin-jdbc:test :plugin-jdbc-postgres:test` passes (core 37 passing; postgres 30 passing, 1 pre-existing `@Disabled`).

### DIFF
--- a/plugin-jdbc/src/main/java/io/kestra/plugin/jdbc/AbstractJdbcTrigger.java
+++ b/plugin-jdbc/src/main/java/io/kestra/plugin/jdbc/AbstractJdbcTrigger.java
@@ -8,6 +8,7 @@ import io.kestra.core.models.property.Property;
 import io.kestra.core.models.tasks.common.FetchType;
 import io.kestra.core.models.triggers.*;
 import io.kestra.core.runners.RunContext;
+import io.swagger.v3.oas.annotations.media.Schema;
 import jakarta.validation.constraints.NotNull;
 import lombok.*;
 import lombok.experimental.SuperBuilder;
@@ -71,7 +72,14 @@ public abstract class AbstractJdbcTrigger extends AbstractTrigger implements Pol
 
     @NotNull
     @Builder.Default
-    protected Property<FetchType> fetchType = Property.ofValue(FetchType.NONE);
+    @Schema(
+        title = "The way you want to fetch the data.",
+        description = """
+            Triggers default to `FETCH`, which loads all rows into memory and exposes them as `{{ trigger.rows }}`. \
+            A trigger fires only when the query returns at least one row, so `NONE` would never fire and must not be used here. \
+            Use `FETCH_ONE` to expose a single row as `{{ trigger.row }}`, or `STORE` to write the rows to internal storage and expose the file URI as `{{ trigger.uri }}`."""
+    )
+    protected Property<FetchType> fetchType = Property.ofValue(FetchType.FETCH);
 
     @Builder.Default
     protected Property<Integer> fetchSize = Property.ofValue(10000);
--- a/plugin-jdbc/src/main/java/io/kestra/plugin/jdbc/AbstractJdbcTrigger.java
+++ b/plugin-jdbc/src/main/java/io/kestra/plugin/jdbc/AbstractJdbcTrigger.java
@@ -8,7 +8,10 @@ import io.kestra.core.models.property.Property;
 import io.kestra.core.models.tasks.common.FetchType;
 import io.kestra.core.models.triggers.*;
 import io.kestra.core.runners.RunContext;
+import io.kestra.core.utils.PebbleUtil;
 import io.swagger.v3.oas.annotations.media.Schema;
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import jakarta.validation.constraints.AssertTrue;
 import jakarta.validation.constraints.NotNull;
 import lombok.*;
 import lombok.experimental.SuperBuilder;
@@ -81,6 +84,16 @@ public abstract class AbstractJdbcTrigger extends AbstractTrigger implements Pol
             Use `FETCH_ONE` to expose a single row as `{{ trigger.row }}`, or `STORE` to write the rows to internal storage and expose the file URI as `{{ trigger.uri }}`."""
     )
     protected Property<FetchType> fetchType = Property.ofValue(FetchType.FETCH);
+
+    @AssertTrue(message = "fetchType NONE is not valid for triggers — the trigger would never fire. Use FETCH, FETCH_ONE, or STORE.")
+    @JsonIgnore
+    boolean isFetchTypeValid() {
+        if (fetchType == null) return true;
+        var expr = fetchType.toString();
+        // Skip validation for dynamic Pebble expressions — those can only be checked at render time
+        if (PebbleUtil.containsOpeningBlockDelimiter(expr)) return true;
+        return !FetchType.NONE.name().equalsIgnoreCase(expr);
+    }
 
     @Builder.Default
     protected Property<Integer> fetchSize = Property.ofValue(10000);
--- a/plugin-jdbc/src/main/java/io/kestra/plugin/jdbc/AbstractJdbcTrigger.java
+++ b/plugin-jdbc/src/main/java/io/kestra/plugin/jdbc/AbstractJdbcTrigger.java
@@ -72,11 +72,12 @@ public abstract class AbstractJdbcTrigger extends AbstractTrigger implements Pol
 
     @NotNull
     @Builder.Default
+    @PluginProperty(group = "main")
     @Schema(
         title = "The way you want to fetch the data.",
         description = """
             Triggers default to `FETCH`, which loads all rows into memory and exposes them as `{{ trigger.rows }}`. \
-            A trigger fires only when the query returns at least one row, so `NONE` would never fire and must not be used here. \
+            A trigger fires only when the query returns at least one row; setting `NONE` would cause the trigger to never fire. \
             Use `FETCH_ONE` to expose a single row as `{{ trigger.row }}`, or `STORE` to write the rows to internal storage and expose the file URI as `{{ trigger.uri }}`."""
     )
     protected Property<FetchType> fetchType = Property.ofValue(FetchType.FETCH);
@@ -118,7 +119,11 @@ public abstract class AbstractJdbcTrigger extends AbstractTrigger implements Pol
         } else if(this.store) {
             return FetchType.STORE;
         }
-        return runContext.render(fetchType).as(FetchType.class).orElseThrow();
+        var resolved = runContext.render(fetchType).as(FetchType.class).orElseThrow();
+        if (resolved == FetchType.NONE) {
+            throw new IllegalArgumentException("fetchType NONE is not valid for triggers — the trigger would never fire. Use FETCH, FETCH_ONE, or STORE.");
+        }
+        return resolved;
     }
 
     protected abstract AbstractJdbcQuery.Output runQuery(RunContext runContext) throws Exception;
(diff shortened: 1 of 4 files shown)


jymaire:
## 🧪 QA Report — Browser-based (Kestra OSS, develop)

**Scope:** PR #930 — JDBC triggers default `fetchType` `NONE` → `FETCH` (so triggers fire out of the box), plus the follow-up guard rejecting `fetchType: NONE` on triggers.

**Environment:** Kestra OSS `develop`, `plugin-jdbc` `1.14.2-SNAPSHOT` shadow JAR built from this branch (`6bcb231c`). Postgres from `docker-compose-ci.yml` (host port `56982`), pre-seeded `pgsql_types` table (2 rows). Triggers polled at `PT10S`; UI used to create/observe executions, Gantt, Logs, Outputs.

### Summary

| # | Flow | Scenario | Result |
|---|------|----------|--------|
| 1 | `trigger_no_fetchtype` | ⭐ Trigger with **no** `fetchType` (the fix) | ✅ SUCCESS — fired, `trigger.rows`/`trigger.size` populated |
| 2 | `trigger_explicit_fetch` | Trigger with explicit `fetchType: FETCH` (non-regression) | ✅ SUCCESS — fired |
| 3 | `trigger_rows_access` | No `fetchType`, accesses `trigger.rows[0].concert_id` | ✅ SUCCESS — nested row fields readable |
| 4 | `trigger_none_guard` | Trigger with `fetchType: NONE` | ✅ Does NOT fire (0 executions) — but see note below |

All happy-path triggers fired well within the 30s timeout. No regressions observed.

---

### Flow 1: `trigger_no_fetchtype` (✅ SUCCESS) — the fix

<details>
<summary>Flow YAML</summary>

```yaml
id: trigger_no_fetchtype
namespace: qa.jdbc
tasks:
  - id: show
    type: io.kestra.plugin.core.debug.Return
    format: "size={{ trigger.size }} rows={{ trigger.rows }}"
triggers:
  - id: watch
    type: io.kestra.plugin.jdbc.postgresql.Trigger
    interval: PT10S
    url: jdbc:postgresql://127.0.0.1:56982/
    username: postgres
    password: pg_passwd
    sql: SELECT concert_id, available, b FROM pgsql_types
```
</details>

**Gantt**
| Task | Status | Duration |
|------|--------|----------|
| `show` | SUCCESS | 0.09s |
| **Total** | **SUCCESS** | **~0.15s** |

**Logs synthesis** — Trigger `watch` evaluated and produced an execution every poll cycle. `Return` rendered:
`size=2 rows=[{"b":"This is a varchar","available":true,"concert_id":1},{"b":"This is a varchar","available":true,"concert_id":2}]`

**Outputs synthesis** (Outputs tab) — `outputs.show.value` = the string above. The execution Overview shows the `watch` trigger with `variables (2 items)` → `rows` (the 2 structured rows) and `size = 2`. **Confirms the trigger fires with the default `FETCH` and exposes `{{ trigger.rows }}` without any explicit `fetchType`.**

---

### Flow 2: `trigger_explicit_fetch` (✅ SUCCESS) — non-regression

<details>
<summary>Flow YAML</summary>

```yaml
id: trigger_explicit_fetch
namespace: qa.jdbc
tasks:
  - id: show
    type: io.kestra.plugin.core.debug.Return
    format: "size={{ trigger.size }}"
triggers:
  - id: watch
    type: io.kestra.plugin.jdbc.postgresql.Trigger
    interval: PT10S
    fetchType: FETCH
    url: jdbc:postgresql://127.0.0.1:56982/
    username: postgres
    password: pg_passwd
    sql: SELECT concert_id, available, b FROM pgsql_types
```
</details>

**Gantt** — `show` SUCCESS. **Logs** — `size=2`. Explicitly setting `fetchType: FETCH` still works exactly as before.

---

### Flow 3: `trigger_rows_access` (✅ SUCCESS) — structured row access

<details>
<summary>Flow YAML</summary>

```yaml
id: trigger_rows_access
namespace: qa.jdbc
tasks:
  - id: return
    type: io.kestra.plugin.core.debug.Return
    format: "first_concert_id={{ trigger.rows[0].concert_id }} b={{ trigger.rows[0].b }} total={{ trigger.size }}"
triggers:
  - id: watch
    type: io.kestra.plugin.jdbc.postgresql.Trigger
    interval: PT10S
    url: jdbc:postgresql://127.0.0.1:56982/
    username: postgres
    password: pg_passwd
    sql: SELECT concert_id, available, b FROM pgsql_types
```
</details>

**Logs** — `first_concert_id=1 b=This is a varchar total=2`. The default-`FETCH` rows are proper structured maps; nested field access works in Pebble.

> Note: the canonical `Trigger.java` `@Example` uses `io.kestra.plugin.core.flow.ForEach` over `{{ trigger.rows }}`. That type did not resolve in this `develop` core build (`Invalid type: io.kestra.plugin.core.flow.ForEach`), so the iterate-rows pattern was validated via direct indexed access instead. This is a core/version detail, unrelated to this PR.

---

### Flow 4: `trigger_none_guard` (✅ does not fire) — `fetchType: NONE`

<details>
<summary>Flow YAML</summary>

```yaml
id: trigger_none_guard
namespace: qa.jdbc
tasks:
  - id: show
    type: io.kestra.plugin.core.debug.Return
    format: "should not run"
triggers:
  - id: watch
    type: io.kestra.plugin.jdbc.postgresql.Trigger
    interval: PT10S
    fetchType: NONE
    url: jdbc:postgresql://127.0.0.1:56982/
    username: postgres
    password: pg_passwd
    sql: SELECT concert_id FROM pgsql_types
```
</details>

**Result** — 0 executions after multiple poll cycles (~25s). Correct: a `NONE` trigger must never fire.

**⚠️ Minor note (not a blocker):** the new `renderFetchType()` guard throws `IllegalArgumentException("fetchType NONE is not valid for triggers …")`, but that message **never reaches the scheduler/worker logs**. The only signal the user sees is the generic:
`WARN s.m.qa.jdbc.trigger_none_guard.watch … No execution found, schedule is blocked since '…'`
— i.e. the same opaque symptom this PR set out to eliminate. The exception is silently swallowed during polling-trigger evaluation. Functionally safe, but the helpful guard message provides no user-visible benefit today. Consider surfacing it (e.g. fail flow validation on `fetchType: NONE` for triggers, or log the cause when trigger evaluation throws).

---

### Environment note (not a PR issue)
At startup the plugin scanner logs `ERROR … Unable to load all plugin classes from '…plugin-jdbc__1_14_2-SNAPSHOT-tests.jar' … NoClassDefFoundError: io/kestra/plugin/jdbc/AbstractJdbcTrigger`. A stray `-tests.jar` was deployed into the plugins dir by the local build tooling; the main JAR loaded fine and all flows worked. Pure local-deploy artifact, unrelated to the PR.

### Timeouts
None. Every happy-path trigger fired within one or two `PT10S` poll cycles.

### Verdict
✅ **PASS.** The core change works as intended: JDBC triggers now fire out of the box with the default `FETCH` and expose `{{ trigger.rows }}`/`{{ trigger.size }}`; explicit `FETCH` is unchanged; `NONE` does not fire. One non-blocking polish item on the `NONE` guard's (invisible) error message.

plugin-jdbc/src/main/java/io/kestra/plugin/jdbc/AbstractJdbcTrigger.java
```diff
@@ -97,7 +75,25 @@ relies on session state persisting on the connection (for example SET search_pat
...
-    protected Property<FetchType> fetchType = Property.ofValue(FetchType.NONE);
+    @PluginProperty(group = "main")
+    @Schema(
+        title = "The way you want to fetch the data.",
+        description = """
+            Triggers default to `FETCH`, which loads all rows into memory and exposes them as `{{ trigger.rows }}`. \
+            A trigger fires only when the query returns at least one row; setting `NONE` would cause the trigger to never fire. \
+            Use `FETCH_ONE` to expose a single row as `{{ trigger.row }}`, or `STORE` to write the rows to internal storage and expose the file URI as `{{ trigger.uri }}`."""
+    )
+    protected Property<FetchType> fetchType = Property.ofValue(FetchType.FETCH);
```
fdelbrayelle, line 86: Maybe there was a reason to have NONE by default but I don't know it :( ... I don't see any breaking change to change the default value here but some users may be surprised.

plugin-jdbc/src/main/java/io/kestra/plugin/jdbc/AbstractJdbcTrigger.java
```diff
@@ -136,7 +132,11 @@ public FetchType renderFetchType(RunContext runContext) throws IllegalVariableEv
         } else if(this.store) {
             return FetchType.STORE;
         }
-        return runContext.render(fetchType).as(FetchType.class).orElseThrow();
+        var resolved = runContext.render(fetchType).as(FetchType.class).orElseThrow();
```
fdelbrayelle:
```suggestion
        var rFetchType = runContext.render(fetchType).as(FetchType.class).orElseThrow();
```

fdelbrayelle (approved):

plugin-jdbc/src/main/java/io/kestra/plugin/jdbc/AbstractJdbcTrigger.java
jymaire: The `NONE` default was inherited from `AbstractJdbcQuery` rather than chosen deliberately for triggers. For a trigger it's actually a non-working default: `AbstractJdbcQuery#run` has no `NONE` arm in its switch, so no rows are read, `size` stays `0`, and `AbstractJdbcTrigger#evaluate` returns `Optional.empty()` — the trigger never fires (the scheduler just reports "schedule is blocked"). So a trigger with only the required fields silently never worked.

Agreed there's no breaking change: explicit `fetchType` values and the deprecated `fetch`/`fetchOne`/`store` booleans are untouched (`renderFetchType()` short-circuits on the booleans first), so only the previously-broken no-`fetchType` case changes behavior — from "never fires" to "fires with FETCH". To remove the surprise entirely I also added a validation-time rejection of `fetchType: NONE` on triggers (fails fast at flow save with a clear message) plus a runtime guard for the dynamic-expression case.

plugin-jdbc/src/main/java/io/kestra/plugin/jdbc/AbstractJdbcTrigger.java
jymaire: Done in 292cf709 — renamed to `rFetchType` to match the `rSql` convention. Thanks!

PR #966, opened 2026-07-20
fix(snowflake): resolve multi-statement count error and closed result set loop crash

### LINKED ISSUE #17481: Snowflake multiple queries fail
### Describe the issue

running on kestra v2.0 (early adopter)

```
id: snowflake
namespace: kestra


pluginDefaults:
  - type: io.kestra.plugin.jdbc.snowflake
    values:
      url: jdbc:snowflake://<my-instance>.snowflakecomputing.com
      account: <my-account>
      username: <my-user>
      privateKey: <my-pk-secret-with-formatting>
      role: <my-role>
      warehouse: <my-warehouse>

tasks:
      - id: snowflake_single_query
        type: io.kestra.plugin.jdbc.snowflake.Query
        sql: |
          SELECT * FROM INGEST_DEV.HUBSPOT.CUSTOMERS LIMIT 2
          
      - id: snowflake_multiple_queries
        type: io.kestra.plugin.jdbc.snowflake.Queries
        sql: SELECT * FROM INGEST_DEV.HUBSPOT.CUSTOMERS LIMIT 2; SELECT * FROM INGEST_DEV.HUBSPOT.CUSTOMERS LIMIT 2;
```

the first task (single query) runs through without any trouble. the second task fails, with no error message (I think the missing error message is a separate problem - other flows have problems with logs as well right now. see error message below in comment).

### Environment

- Kestra Version: v2.0.0-rc5

narendra-003:
This PR resolves two critical runtime errors encountered when executing multi-statement SQL strings in Snowflake. It completely closes kestra-io/kestra#17481.

1. **The Statement Count Error**: Fixed the driver rejecting multi-query strings with the error `Actual statement count did not match the desired statement count 1`. This was resolved by appending `"MULTI_STATEMENT_COUNT", "0"` to the connection properties in the Snowflake-specific `Queries.java` task.
2. **The Closed Resource Error**: Fixed the multi-result loop crash where the exception `Result set has been closed` was thrown. When `stmt.getMoreResults(Statement.CLOSE_CURRENT_RESULT)` finishes consuming all data streams, the Snowflake/Hikari proxy objects retain a non-null state reference where `isClosed() == true`. Added an explicit `rs.isClosed()` validation paired with the `updateCount == -1` termination condition at the top of the shared `AbstractJdbcQueries` loop to ensure a clean, safe termination before reading an invalidated resource.

---

This flow successfully executes two successive queries against a Snowflake data warehouse, which previously failed due to the driver restricting query counts or crashing inside the loop.

```yaml
id: snowflake
namespace: kestra

pluginDefaults:
  - type: io.kestra.plugin.jdbc.snowflake
    values:
      url: jdbc:snowflake://<my-instance>.snowflakecomputing.com
      account: <my-account>
      username: <my-user>
      privateKey: <my-pk-secret-with-formatting>
      role: <my-role>
      warehouse: <my-warehouse>

tasks:
      - id: snowflake_multiple_queries
        type: io.kestra.plugin.jdbc.snowflake.Queries
        sql: <SQL_QUERY>; <SQL_QUERY>;
```

---

### Setup Instructions

- A working Snowflake cloud data warehouse account is required to test this configuration natively. 
- Ensure that the execution target context has permission to execute metadata commands such as `SHOW ROLES`.

---

### Contributor Checklist ✅


fdelbrayelle:
## Code Review — Staff Engineer + Security

**Verdict: Request changes**

Fix targets a confirmed bug (kestra#17481: `SnowflakeSQLException` statement-count mismatch). Setting `MULTI_STATEMENT_COUNT=0` on the Snowflake connection is correctly scoped to `Queries.java` only.

### Findings

🟠 **No executing test coverage**
The shared loop in `AbstractJdbcQueries.extractResultsFromResultSet` (multi-statement branch) is reached by Redshift/Snowflake/Sybase and MySQL/MariaDB with `allowMultiQueries=true`. `SnowflakeQueriesTest`, `RedshiftQueriesTest`, `SybaseQueriesTest` are all `@Disabled`, and `QueriesMysqlTest` doesn't set `allowMultiQueries=true` — so this exact code path has zero CI coverage before or after this change. Please add a test proving the loop terminates cleanly on a multi-statement batch (a driver-agnostic unit test simulating a non-null-but-closed `ResultSet` would work if a live Snowflake gate isn't feasible).

🟡 **Comment style**
`// 1. CRITICAL: True JDBC check for end of all results` — banner/narration style. Suggest a plain one-liner explaining the WHY (some drivers return a non-null but already-closed `ResultSet` at end of stream).

🟡 **Termination guard may be narrower than the root cause**
`if ((rs == null || rs.isClosed()) && updateCount == -1) break;` only catches the terminal case. If the same closed-`ResultSet` quirk can occur mid-batch (right after a DML statement, not just at the tail), `updateCount` won't be `-1` there, the guard won't fire, and the original crash could reproduce mid-stream instead. Consider decoupling the closed-check from `updateCount == -1`.

🟡 **PR description scope**
Framed as Snowflake-only, but the second change touches a method shared by Redshift/Sybase/MySQL — worth flagging for reviewers/QA doing regression thinking.

### Security
No blocking issues. Note: raising `MULTI_STATEMENT_COUNT` to unbounded removes an incidental defense against stacked-query injection for Snowflake specifically, but this matches Redshift/Sybase's existing unbounded behavior on the same task family — not a regression, just worth being aware there's no real statement-count control here.

jymaire:
Adding one finding on top of @fdelbrayelle's review.

🟡 **Dead branch left behind in `AbstractJdbcQueries.extractResultsFromResultSet`**

The new guard at line 222 makes the existing termination check at line 260 unreachable:

```java
// line 222 — new
if ((rs == null || rs.isClosed()) && updateCount == -1) {
    break;
}
if (rs != null) {
    // ...
} else if (updateCount == -1) {   // line 260 — now unreachable
    // End of results
    break;
}
```

Reaching line 260 requires `rs == null`, and `rs == null && updateCount == -1` has already broken out at line 222. So the `else if` can only be entered with `rs == null && updateCount != -1`, where its condition is false by construction.

Not a behavior change, but it leaves two termination checks for one condition in a loop that's shared by Redshift/Snowflake/Sybase and MySQL/MariaDB with `allowMultiQueries=true`. Worth consolidating into a single check so the next reader doesn't have to work out which one actually fires.

This also disappears on its own if the guard is reworked as @fdelbrayelle suggested (decoupling the closed-check from `updateCount == -1`) — one option there is to drive the loop off the boolean returned by `stmt.execute()` / `stmt.getMoreResults()` instead of re-calling `getResultSet()` after the results are exhausted, which avoids touching an invalidated `ResultSet` at all. `stmt.execute()`'s return value is currently discarded at line 99.

Malaydewangan09:
Hey @narendra-003 👋, are there any updates on this?

MilosPaunovic:
Another ping for this one @narendra-003?

narendra-003:
Hi @MilosPaunovic,
Working on the requested changes — test coverage, the mid-batch termination guard fix, and the comment/description cleanup. Pushing shortly. Sorry for the delay in response.

MilosPaunovic:
Hey @narendra-003, are there any updates on this?

narendra-003:
Hi @MilosPaunovic, could you please unassign this issue from me for now? Sorry for the inconvenience. I’m currently busy with other priorities, but I’d love to contribute in the future. Thanks!

PR #972, opened 2026-08-03
feat(clickhouse): add embedded ChDB task for lightweight data transforms

### LINKED ISSUE #481: Add ClickHouse chDB plugin to provide embedded DB for lightweight data transformations
### Feature description

Like DuckDB, chDB is an in-memory OLAP engine. It has some advantages, especially a [high number of format integration](https://clickhouse.com/docs/en/interfaces/formats).

- Contribute to the java binding (first iteration can be found here, not sure if relevant or better to start from scratch https://github.com/chdb-io/chdb-java)
- https://github.com/chdb-io/chdb/blob/main/bindings.md

rishi-jat:

Adds `io.kestra.plugin.jdbc.clickhouse.ChDB`, an embedded ClickHouse task backed by `clickhouse-local` for lightweight data transformations without requiring a ClickHouse server.

- Adds `query` and optional `outputFormat` properties as proposed in #481.
- Stores results as Ion by default (and for `Pretty*` formats).
- Stores file-oriented formats (e.g. `CSVWithNames`, `Parquet`) using matching file extensions.

Closes #481

---

```yaml
id: chdb_demo
namespace: company.team

tasks:
  - id: ion
    type: io.kestra.plugin.jdbc.clickhouse.ChDB
    query: |
      SELECT number AS n
      FROM system.numbers
      LIMIT 5

  - id: csv
    type: io.kestra.plugin.jdbc.clickhouse.ChDB
    query: |
      SELECT number AS n
      FROM system.numbers
      LIMIT 5
    outputFormat: CSVWithNames
```
- Unit tests: format → Ion / extension mapping
- Integration tests: Docker + clickhouse-local (when Docker is available)

---

### Setup Instructions
Setup Instructions
- Docker required for the default task runner (clickhouse/clickhouse-server:latest)
- Or use a Process runner with clickhouse-local on PATH

───

### Contributor Checklist ✅

### DIFF
--- a/plugin-jdbc-clickhouse/src/main/java/io/kestra/plugin/jdbc/clickhouse/ChDB.java
+++ b/plugin-jdbc-clickhouse/src/main/java/io/kestra/plugin/jdbc/clickhouse/ChDB.java
@@ -214,11 +214,12 @@ public class ChDB extends Task implements RunnableTask<ChDB.Output>, NamespaceFi
 
     @Override
     public Output run(RunContext runContext) throws Exception {
-        String rQuery = normalizeAndValidateQuery(
-            runContext.render(this.query).as(String.class).orElseThrow()
+        var rQuery = normalizeAndValidateQuery(
+            runContext.render(this.query).as(String.class)
+                .orElseThrow(() -> new IllegalArgumentException("Property 'query' must be set"))
         );
 
-        String rOutputFormat = this.outputFormat == null
+        var rOutputFormat = this.outputFormat == null
             ? null
             : runContext.render(this.outputFormat).as(String.class).orElse(null);
 
@@ -231,25 +232,25 @@ public class ChDB extends Task implements RunnableTask<ChDB.Output>, NamespaceFi
             }
         }
 
-        boolean storeAsIon = ChDBOutputFormats.shouldStoreAsIon(rOutputFormat);
+        var storeAsIon = ChDBOutputFormats.shouldStoreAsIon(rOutputFormat);
         // Defensive validation: `clickHouseFormat` is interpolated into a shell command.
-        String clickHouseFormat = ChDBOutputFormats.requireSafeFormat(
+        var clickHouseFormat = ChDBOutputFormats.requireSafeFormat(
             ChDBOutputFormats.effectiveClickHouseFormat(rOutputFormat)
         );
 
         // Intermediate file written by clickhouse-local inside the working directory.
         // Ion path uses JSONEachRow then converts; file formats keep their dedicated extension.
-        String rawFileName = storeAsIon ? "result.jsonl" : ChDBOutputFormats.outputFileName(rOutputFormat);
+        var rawFileName = storeAsIon ? "result.jsonl" : ChDBOutputFormats.outputFileName(rOutputFormat);
 
-        Path workingDir = runContext.workingDir().path();
-        Path queryFile = workingDir.resolve("query.sql");
+        var workingDir = runContext.workingDir().path();
+        var queryFile = workingDir.resolve("query.sql");
         Files.writeString(queryFile, rQuery, StandardCharsets.UTF_8);
 
         // Prefer shell redirection over INTO OUTFILE so we do not depend on allow_file_output settings.
         // --format=VALUE avoids shell word-splitting for multi-word format names.
-        String shellCommand = clickhouseLocalShellCommand(clickHouseFormat, rawFileName, storeAsIon);
+        var shellCommand = clickhouseLocalShellCommand(clickHouseFormat, rawFileName, storeAsIon);
 
-        ScriptOutput scriptOutput = new CommandsWrapper(runContext)
+        var scriptOutput = new CommandsWrapper(runContext)
             .withWarningOnStdErr(true)
             .withTaskRunner(this.taskRunner)
             .withContainerImage(this.containerImage)
@@ -267,19 +268,19 @@ public class ChDB extends Task implements RunnableTask<ChDB.Output>, NamespaceFi
             );
         }
 
-        Map<String, URI> outputFiles = scriptOutput.getOutputFiles();
+        var outputFiles = scriptOutput.getOutputFiles();
         if (outputFiles == null || !outputFiles.containsKey(rawFileName)) {
             throw new IllegalStateException(
                 "clickhouse-local did not produce expected output file '" + rawFileName + "'"
             );
         }
 
-        URI rawUri = outputFiles.get(rawFileName);
+        var rawUri = outputFiles.get(rawFileName);
         Long rowCount = null;
         URI resultUri;
 
         if (storeAsIon) {
-            Path ionPath = runContext.workingDir().createTempFile(ChDBOutputFormats.DEFAULT_ION_EXTENSION);
+            var ionPath = runContext.workingDir().createTempFile(ChDBOutputFormats.DEFAULT_ION_EXTENSION);
             rowCount = convertJsonEachRowToIon(runContext, rawUri, ionPath);
             resultUri = runContext.storage().putFile(ionPath.toFile());
         } else {
@@ -300,16 +301,16 @@ public class ChDB extends Task implements RunnableTask<ChDB.Output>, NamespaceFi
      * @return number of rows written
      */
     private static long convertJsonEachRowToIon(RunContext runContext, URI jsonEachRowUri, Path ionPath) throws Exception {
-        long rows = 0L;
-        File ionFile = ionPath.toFile();
+        var rows = 0L;
+        var ionFile = ionPath.toFile();
 
         try (
-            InputStream inputStream = runContext.storage().getFile(jsonEachRowUri);
-            BufferedReader reader = new BufferedReader(
+            var inputStream = runContext.storage().getFile(jsonEachRowUri);
+            var reader = new BufferedReader(
                 new InputStreamReader(inputStream, StandardCharsets.UTF_8),
                 FileSerde.BUFFER_SIZE
             );
-            BufferedOutputStream output = new BufferedOutputStream(
+            var output = new BufferedOutputStream(
                 new FileOutputStream(ionFile),
                 FileSerde.BUFFER_SIZE
             )
@@ -319,7 +320,7 @@ public class ChDB extends Task implements RunnableTask<ChDB.Output>, NamespaceFi
                 if (line.isBlank()) {
                     continue;
                 }
-                Map<String, Object> row = JSON_MAPPER.readValue(line, MAP_TYPE);
+                var row = JSON_MAPPER.readValue(line, MAP_TYPE);
                 FileSerde.write(output, row);
                 rows++;
             }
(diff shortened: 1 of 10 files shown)


plugin-jdbc-clickhouse/src/main/java/io/kestra/plugin/jdbc/clickhouse/ChDB.java
```diff
@@ -0,0 +1,325 @@
...
+        // Ion path uses JSONEachRow then converts; file formats keep their dedicated extension.
+        String rawFileName = storeAsIon ? "result.jsonl" : ChDBOutputFormats.outputFileName(userFormat);
+
+        Path workingDir = runContext.workingDir().path();
+        Path queryFile = workingDir.resolve("query.sql");
+        Files.writeString(queryFile, renderedQuery, StandardCharsets.UTF_8);
+
+        // Prefer shell redirection over INTO OUTFILE so we do not depend on allow_file_output settings.
+        // --format=VALUE avoids shell word-splitting for multi-word format names.
+        String shellCommand = "clickhouse-local --queries-file=query.sql --format=" + clickHouseFormat
```
Malaydewangan09, line 242: `clickHouseFormat` is user input (`outputFormat`) and goes straight into `/bin/sh -c` here, so `outputFormat: "CSV; <anything>"` runs arbitrary commands. The query is safe (written to a file), this isn't. ClickHouse format names are alphanumeric, so validating against `[A-Za-z0-9]+` (or an allowlist) before building the command closes it.

plugin-jdbc-clickhouse/src/main/java/io/kestra/plugin/jdbc/clickhouse/ChDBOutputFormats.java
```diff
@@ -0,0 +1,122 @@
...
+        // Fall back to a sanitized format name so the artifact remains identifiable.
+        String sanitized = format.replaceAll("[^a-z0-9]+", "_");
+        return "." + (sanitized.isBlank() ? "out" : sanitized);
+    }
+
+    static String outputFileName(String outputFormat) {
+        return "result" + fileExtension(outputFormat);
+    }
+
+    static Optional<String> normalizeFormat(String outputFormat) {
```
Malaydewangan09: `normalizeFormat` isn't called anywhere in the module. Dead code, safe to remove.

plugin-jdbc-clickhouse/src/main/java/io/kestra/plugin/jdbc/clickhouse/ChDB.java
```diff
@@ -0,0 +1,325 @@
...
+            title = "URI of the query result in Kestra internal storage",
+            description = "Ion file when no format / Pretty* is used; otherwise a file with a format-specific extension."
+        )
+        private final URI uri;
+
+        @Schema(
+            title = "Number of rows written",
+            description = "Populated when the result is stored as Ion (default / Pretty* formats). Null for raw file formats."
+        )
+        private final Long size;
```
Malaydewangan09: `size` holds the row count (from `convertJsonEachRowToIon`), not a byte size, but it reads as bytes. `rowCount` would be clearer, and it's easier to rename before this ships.

plugin-jdbc-clickhouse/src/main/java/io/kestra/plugin/jdbc/clickhouse/ChDB.java
```diff
@@ -0,0 +1,345 @@
...
+    private NamespaceFiles namespaceFiles;
+
+    @PluginProperty(group = "source")
+    private Object inputFiles;
+
+    @Override
+    public Output run(RunContext runContext) throws Exception {
+        String renderedQuery = runContext.render(this.query).as(String.class).orElseThrow().strip();
+        // Strip trailing semicolons; FORMAT / INTO OUTFILE are controlled by this task.
+        renderedQuery = renderedQuery.replaceAll(";\\s*$", "");
```
Malaydewangan09: Only the trailing `;` is stripped, so multiple `;`-separated statements still pass through and clickhouse-local runs all of them, keeping only the last output. The schema says single statement, so either enforce that or note the behavior.

plugin-jdbc-clickhouse/src/main/java/io/kestra/plugin/jdbc/clickhouse/ChDB.java
```diff
@@ -0,0 +1,345 @@
...
+
+        if (renderedQuery.isBlank()) {
+            throw new IllegalArgumentException("Property 'query' must not be blank");
+        }
+        // FORMAT as a clause (not format() / formatDateTime()), and INTO OUTFILE, conflict with
+        // the task-controlled --format and shell redirection.
+        if (renderedQuery.matches("(?is).*\\bFORMAT\\s+[A-Za-z0-9]+.*")) {
```
Malaydewangan09: The FORMAT guard is a plain regex, so it also rejects queries where the word shows up in a string literal or alias, like `SELECT 'FORMAT JSON' AS note`. Fine as a heuristic, but worth a comment saying so, since it will trip someone up eventually.

plugin-jdbc-clickhouse/src/main/java/io/kestra/plugin/jdbc/clickhouse/ChDB.java
```diff
@@ -0,0 +1,345 @@
...
+            """
+    )
+    @PluginProperty(group = "main")
+    private Property<String> outputFormat;
+
+    @Schema(
+        title = "Additional environment variables for the process / container"
+    )
+    @PluginProperty(dynamic = true, group = "execution")
+    protected Map<String, String> env;
```
Malaydewangan09: `env` is `protected` while the other fields here are `private`. Looks copied from `ClickHouseLocalCLI`, no reason for it to be `protected` here.

plugin-jdbc-clickhouse/src/main/java/io/kestra/plugin/jdbc/clickhouse/ChDB.java
```diff
@@ -0,0 +1,345 @@
...
+            );
+        }
+
+        URI rawUri = outputFiles.get(rawFileName);
+        Long rowCount = null;
+        URI resultUri;
+
+        if (storeAsIon) {
+            Path ionPath = runContext.workingDir().createTempFile(ChDBOutputFormats.DEFAULT_ION_EXTENSION);
+            rowCount = convertJsonEachRowToIon(runContext, rawUri, ionPath);
```
Malaydewangan09, line 284: The Ion path goes JSONEachRow -> Map -> Ion. ClickHouse defaults `output_format_json_quote_64bit_integers=1`, so `Int64`/`UInt64` come back as JSON strings, not numbers. A big-int column stored through ChDB then won't match the same column from the JDBC `Query` task. Not a blocker, but please note it in the `outputFormat` doc, and consider setting the `output_format_json_*` settings for closer type parity.

plugin-jdbc-clickhouse/src/main/java/io/kestra/plugin/jdbc/clickhouse/ChDB.java
```diff
@@ -0,0 +1,345 @@
...
+     * @return number of rows written
+     */
+    private static long convertJsonEachRowToIon(RunContext runContext, URI jsonEachRowUri, Path ionPath) throws Exception {
+        long rows = 0L;
+        File ionFile = ionPath.toFile();
+
+        try (
+            InputStream inputStream = runContext.storage().getFile(jsonEachRowUri);
+            BufferedReader reader = new BufferedReader(
+                new java.io.InputStreamReader(inputStream, StandardCharsets.UTF_8),
```
Malaydewangan09: `new java.io.InputStreamReader(...)` is fully qualified while the rest of the file uses imports. Add the import to match.

Malaydewangan09 (commented):
Latest changes look good. The validation plus the FORMAT / INTO OUTFILE guards close the injection concern, and my earlier notes (size to rowCount, normalizeFormat removal) are resolved.

One design question before merge: this runs the `clickhouse-local` binary, the same one the existing `ClickHouseLocalCLI` task already uses. The Ion and extension handling is a genuine add, but we need a clear 'when to use ChDB vs ClickHouseLocalCLI' line, and the `ChDB` name reads as libchdb when it is really clickhouse-local. The type name is public API and hard to change later, so worth confirming it is what we want.

A few inline notes below.

fdelbrayelle (changes requested):
## Code Review — Staff Engineer + Security review

**Verdict: Request changes**

### Blocking

🟠 **Bare `.orElseThrow()` on mandatory `query` render** — `ChDB.java:236`
```java
String rQuery = normalizeAndValidateQuery(
    runContext.render(this.query).as(String.class).orElseThrow()
);
```
If rendering yields empty, this throws an opaque `NoSuchElementException` with no context. `@NotNull` guards the property, not the rendered value.
Fix: `.orElseThrow(() -> new IllegalArgumentException("Property 'query' must be set"))`.

🟠 **Backslash-escaped quotes break the multi-statement guard** — `ChDB.java` `hasAdditionalStatement()` (~L382-460)
The tokenizer only recognizes doubled `''` as an escaped quote. ClickHouse also supports backslash escaping (`\'`) by default. A query like:
```sql
SELECT 'Kestra\'s ChDB' AS name; DROP TABLE foo
```
gets misclassified — a legitimate single statement containing `it's`/`O'Brien` could trip the guard, or a real trailing statement after an escaped quote could slip past it undetected. Not covered by `ChDBTest.allowsSemicolonInsideLiteralsCommentsAndTrailing`.
Fix: handle `\` escaping in the state machine (skip the char after `\` while `inSingle`), or explicitly document the limitation in the `query` `@Schema`.

🟠 **`var` not used for local variables** — `ChDB.java:235-306`, `ChDBTest.java`
Nearly every new local (`rQuery`, `rOutputFormat`, `storeAsIon`, `clickHouseFormat`, `rawFileName`, `workingDir`, `queryFile`, `shellCommand`, `scriptOutput`, `outputFiles`, `rawUri`) uses an explicit type instead of `var`, against project convention and against the sibling `ClickHouseLocalCLI.java` in the same package, which already uses `var`.
Fix: switch to `var` wherever legally inferable.

### Non-blocking / maintainer discretion

🟡 DRY — `CommandsWrapper` wiring (taskRunner/containerImage/env/namespaceFiles/inputFiles) is duplicated verbatim between `ChDB` and `ClickHouseLocalCLI`. Consider a shared abstract base if a third `clickhouse-local`-backed task appears.

🟡 No `fetchType` (`FETCH_ONE`/`FETCH`/`STORE`) on `ChDB`, unlike sibling `Query`/`Queries` tasks — always behaves STORE-like. Safe default, but inconsistent UX for small ad-hoc queries. Worth a design note if intentional.

🟡 Per-row JSON parse failure (`ChDB.java:340`) isn't wrapped — a malformed JSONEachRow line throws a raw Jackson exception with no row context.

🟢 No `io.kestra.plugin.jdbc.clickhouse.md` doc page exists yet for this plugin — opportunistic addition since this task changes its capabilities.

### Security
No injection issues found: `outputFormat` is strictly validated (`[A-Za-z0-9]+`) before shell interpolation, and the query text goes through `--queries-file`, never through the shell string directly.

---
*Automated review via kestra-plugin-code-reviewer agent.*

Malaydewangan09:
Hey @rishi-jat  👋, are there any updates on this?

Malaydewangan09 (approved):

fdelbrayelle (approved):

PR #990, opened 2026-08-27
Add Query and Queries plugin examples for Apache DataFusion using Arrow Flight SQL plugin

### LINKED ISSUE #522: Add Query and Queries plugin examples for Apache Data Fusion using Arrow Flight SQL plugin
### Feature description

Trendy Rust project and has Java bindings: https://github.com/datafusion-contrib/datafusion-java

nancysangani:
## What
Adds Apache DataFusion examples to both the `Query` and `Queries` tasks in `plugin-jdbc-arrow-flight`, alongside the existing generic and Dremio examples.

## Why
Closes #522. DataFusion (Rust, with Java bindings via datafusion-java) can be exposed as a Flight SQL server — e.g. via `datafusion-flight-sql-server` — which listens on `localhost:50051` by default. This plugin already supports any Flight SQL-compliant server, but had no example showing DataFusion specifically.

## Changes
- `Query.java`: added a `@Example` connecting to a DataFusion Flight SQL
  server on `localhost:50051`
- `Queries.java`: added the equivalent multi-statement example

No behavioral or dependency changes — documentation/examples only.

### DIFF
--- a/plugin-jdbc-arrow-flight/src/main/java/io/kestra/plugin/jdbc/arrowflight/Queries.java
+++ b/plugin-jdbc-arrow-flight/src/main/java/io/kestra/plugin/jdbc/arrowflight/Queries.java
@@ -47,6 +47,21 @@ import java.time.ZoneId;
                     sql: SELECT * FROM employee; SELECT * FROM laptop;
                     fetchType: FETCH
                 """
+        ),
+        @Example(
+            title = "Send multiple SQL queries to an Apache DataFusion server exposed via Arrow Flight SQL and fetch rows as output.",
+            full = true,
+            code = """
+                id: arrow_flight_sql_datafusion_queries
+                namespace: company.team
+
+                tasks:
+                  - id: query
+                    type: io.kestra.plugin.jdbc.arrowflight.Queries
+                    url: jdbc:arrow-flight-sql://localhost:50051/?useEncryption=false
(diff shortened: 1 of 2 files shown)


fdelbrayelle:
## Code Review — APPROVE ✅

**Scope check against issue #522 discussion:** DataFusion has no native JDBC support, so the ask was pivoted (per @anna-geller's comment) to a plugin *example* showing DataFusion via the existing Arrow Flight SQL plugin. This PR matches that agreed scope exactly — no scope creep, no behavioral changes.

**Findings:** 0 blocking, 0 security, 0 performance issues. Pure doc/example addition (`@Example` on `Query.java` and `Queries.java`), both using `full = true`, proper `id`/`namespace`, consistent with existing examples in the file.

**Non-blocking suggestion:** unlike the existing generic/Dremio examples in these files, the new DataFusion examples omit `username`/`password`. Presumably intentional since a local `datafusion-flight-sql-server` runs without auth by default — but a one-line comment or title note (e.g. "...running without authentication") would make that explicit rather than leaving readers to wonder if credentials were forgotten.

_(Pre-existing, unrelated to this PR: the two existing examples in `Query.java` both reuse the flow id `arrow_flight_sql_query` — a latent duplication, out of scope here.)_

fdelbrayelle (approved):

PR #992, opened 2026-08-28
fix: parse fractional seconds in JDBC-format datetime strings

### LINKED ISSUE #991: JDBC-format datetime strings with fractional seconds fail to parse (Batch / Query parameters)
### Describe the issue

`AbstractCellConverter.parseStringToTimestamp(String)` cannot parse a datetime string that carries a fractional-seconds part. Any such value hits a `DateTimeParseException` (surfaced as *"Unable to transform data..."*) on the `Batch` insert path and on the `Query` `parameters` / after-SQL paths, for **every** database module (the method lives in the shared `plugin-jdbc` base).

This includes the exact format produced by `java.sql.Timestamp.toString()` — `"yyyy-MM-dd HH:mm:ss.f[fffffffff]"` — so feeding a timestamp value back in as a string (a very common shape for `Batch` inputs sourced from an earlier `Query`) fails.

### Root cause

`parseStringToTimestamp` builds a single formatter by chaining `appendOptional(...)` sections:

```java
new DateTimeFormatterBuilder()
    .appendOptional(DateTimeFormatter.ISO_ZONED_DATE_TIME)
    .appendOptional(DateTimeFormatter.ISO_OFFSET_DATE_TIME)
    .appendOptional(DateTimeFormatter.ISO_LOCAL_DATE_TIME)
    .appendOptional(DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm:ss"))
    .appendOptional(DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm:ss.S")) // With milliseconds
    .appendOptional(DateTimeFormatter.ISO_LOCAL_DATE)
    .toFormatter();
```

`appendOptional` sections are applied **sequentially**, not as alternatives. For `"2019-10-30 12:00:00.123"` the `"yyyy-MM-dd HH:mm:ss"` section consumes the first 19 characters, then the next section (`...ss.S`) tries to match *from index 19* — it expects another full `yyyy-MM-dd...` there, fails, and (being optional) is skipped. Parsing then ends with unparsed text remaining → `DateTimeParseException`.

So the `...ss.S` section is effectively dead code, and even if it were reachable, `S` matches exactly one fraction digit.

### Reproduction

```java
// fails: "Text '2019-10-30 12:00:00.123' could not be parsed, unparsed text found at index 19"
new SomeCellConverter(ZoneId.of("UTC"))
    .addPreparedStatementValue(ps, timestampParam, "2019-10-30 12:00:00.123", 1, null);
```

`.1`, `.12`, `.123`, `.123456789` all fail; `"2019-10-30 12:00:00"` (no fraction) works.

### Expected

Datetime strings with a 1–9 digit fractional-seconds part parse successfully, matching `java.sql.Timestamp.toString()` output.

karthikchundi-commits:

closes #991

`AbstractCellConverter.parseStringToTimestamp(String)` could not parse a datetime string with a fractional-seconds part. On the `Batch` insert path and the `Query` `parameters` / after-SQL paths, a value such as `2019-10-30 12:00:00.123` failed with `DateTimeParseException`. This is the shared base method, so every database module was affected, and the failing format is exactly what `java.sql.Timestamp.toString()` produces.

**Root cause:** the method built one formatter by chaining `appendOptional(...)` sections. Those sections are applied sequentially, not as alternatives — once `"yyyy-MM-dd HH:mm:ss"` consumed the datetime, the following `"yyyy-MM-dd HH:mm:ss.S"` section tried to match from the leftover `.123` (still expecting a full `yyyy-MM-dd` prefix), failed, and was skipped as optional. The `...ss.S` section was effectively unreachable, and it only allowed a single fraction digit anyway.

**Fix:** replace the single chained formatter with a list of formatters, each attempted independently against the whole input. The JDBC formatter now accepts an optional 1–9 digit fractional-seconds part via `appendFraction(NANO_OF_SECOND, 1, 9, true)`. ISO handling and the plain-date fallback are unchanged.

Regression cases added to `AbstractCellConverterTest` for `.1`, `.123`, and `.123456789`.

`./gradlew :plugin-jdbc:test` — 42/42 pass (was 39, +3 new). The new cases fail on `main` and pass with this change; existing datetime-parsing tests are unaffected.

```yaml
id: batch_fractional_timestamp
namespace: company.team
tasks:
  - id: gen
    type: io.kestra.plugin.core.debug.Return
    format: |
      { "id": 1, "ts": "2019-10-30 12:00:00.123" }
  - id: to_ion
    type: io.kestra.plugin.serdes.json.JsonToIon
    from: "{{ outputs.gen.value }}"
  - id: insert
    type: io.kestra.plugin.jdbc.postgresql.Batch
    from: "{{ outputs.to_ion.uri }}"
    url: jdbc:postgresql://localhost:5432/kestra
    username: kestra
    password: k3str4
    sql: INSERT INTO events (id, ts) VALUES (?, ?)
```

Before this change the `insert` task fails to transform the `ts` string; after it, the row is inserted with the sub-second precision preserved.

---

### Contributor Checklist ✅

<sub>AI-assisted: investigation, fix, and tests were prepared with Claude Code under my direction and reviewed by me.</sub>

### DIFF
--- a/plugin-jdbc/src/main/java/io/kestra/plugin/jdbc/AbstractCellConverter.java
+++ b/plugin-jdbc/src/main/java/io/kestra/plugin/jdbc/AbstractCellConverter.java
@@ -12,6 +12,7 @@ import java.time.*;
 import java.time.format.DateTimeFormatter;
 import java.time.format.DateTimeFormatterBuilder;
 import java.time.format.DateTimeParseException;
+import java.time.temporal.ChronoField;
 import java.time.temporal.TemporalAccessor;
 import java.util.*;
 
@@ -236,36 +237,54 @@ public abstract class AbstractCellConverter {
         throw addPreparedStatementException(parameterType, index, value, null);
     }
 
+    /**
+     * Accepted string date/time formats, tried in order. Each is attempted against the whole
+     * input independently - unlike a single builder with chained {@code appendOptional(...)}
+     * sections, where an earlier section consuming the date/time part leaves a later section
+     * (e.g. the fractional-seconds one) unable to match the remaining input.
+     */
+    private static final List<DateTimeFormatter> TIMESTAMP_FORMATTERS = List.of(
+        // Standard ISO formats (T separator)
+        DateTimeFormatter.ISO_ZONED_DATE_TIME,
+        DateTimeFormatter.ISO_OFFSET_DATE_TIME,
+        DateTimeFormatter.ISO_LOCAL_DATE_TIME,
+        // JDBC / java.sql.Timestamp.toString() format (space separator), with an optional
+        // fractional-seconds part of 1 to 9 digits ("2019-10-30 12:00:00[.123456789]")
+        new DateTimeFormatterBuilder()
+            .appendPattern("yyyy-MM-dd HH:mm:ss")
+            .optionalStart()
+            .appendFraction(ChronoField.NANO_OF_SECOND, 1, 9, true)
+            .optionalEnd()
+            .toFormatter(),
+        // fallback to just Date
+        DateTimeFormatter.ISO_LOCAL_DATE
+    );
+
     protected Timestamp parseStringToTimestamp(String value) {
-        DateTimeFormatter formatter = new DateTimeFormatterBuilder()
-            // for Standard ISO formats (T separator)
-            .appendOptional(DateTimeFormatter.ISO_ZONED_DATE_TIME)
-            .appendOptional(DateTimeFormatter.ISO_OFFSET_DATE_TIME)
-            .appendOptional(DateTimeFormatter.ISO_LOCAL_DATE_TIME)
+        for (DateTimeFormatter formatter : TIMESTAMP_FORMATTERS) {
+            TemporalAccessor temporal;
+            try {
+                temporal = formatter.parseBest(
+                    value,
+                    ZonedDateTime::from,
+                    LocalDateTime::from,
+                    LocalDate::from
+                );
+            } catch (DateTimeParseException ignored) {
+                continue;
+            }
 
-            // JDBC format (Space separator): "yyyy-MM-dd HH:mm:ss"
-            .appendOptional(DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm:ss"))
-            .appendOptional(DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm:ss.S")) // With milliseconds
+            return switch (temporal) {
+                case ZonedDateTime zdt -> Timestamp.from(zdt.toInstant());
+                case LocalDateTime ldt -> Timestamp.valueOf(ldt);
+                case LocalDate ld      -> Timestamp.valueOf(ld.atStartOfDay());
+                default                -> throw new IllegalArgumentException(
+                    "Unsupported date format: " + value + " (parsed as: " + temporal.getClass().getSimpleName() + ")"
+                );
+            };
+        }
 
-            // fallback to just Date
-            .appendOptional(DateTimeFormatter.ISO_LOCAL_DATE)
-            .toFormatter();
-
-        TemporalAccessor temporal = formatter.parseBest(
-            value,
-            ZonedDateTime::from,
-            LocalDateTime::from,
-            LocalDate::from
-        );
-
-        return switch (temporal) {
-            case ZonedDateTime zdt -> Timestamp.from(zdt.toInstant());
-            case LocalDateTime ldt -> Timestamp.valueOf(ldt);
-            case LocalDate ld      -> Timestamp.valueOf(ld.atStartOfDay());
-            default                -> throw new IllegalArgumentException(
-                "Unsupported date format: " + value + " (parsed as: " + temporal.getClass().getSimpleName() + ")"
-            );
-        };
+        throw new IllegalArgumentException("Unsupported date format: " + value);
     }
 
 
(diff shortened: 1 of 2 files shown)


plugin-jdbc/src/main/java/io/kestra/plugin/jdbc/AbstractCellConverter.java
```diff
@@ -236,36 +237,54 @@ else if (value instanceof String) {
...
-            .appendOptional(DateTimeFormatter.ISO_LOCAL_DATE)
-            .toFormatter();
-
-        TemporalAccessor temporal = formatter.parseBest(
-            value,
-            ZonedDateTime::from,
-            LocalDateTime::from,
-            LocalDate::from
-        );
+        for (DateTimeFormatter formatter : TIMESTAMP_FORMATTERS) {
```
fdelbrayelle, line 297: 🟡 [Performance] Exception-driven formatter matching in a hot loop
Problem: parseStringToTimestamp is invoked per-cell for every String-typed date/time JDBC bind parameter (Batch insert of a CSV/ION file with millions of rows is the exact use case this PR's description calls out). The new loop relies on catching DateTimeParseException per failed attempt, and each exception pays JVM stack-trace-fill cost. For the format this PR specifically targets (the JDBC/Timestamp.toString() space-separated format, listed 4th in TIMESTAMP_FORMATTERS), every matching value now throws and discards 3 exceptions (ISO_ZONED_DATE_TIME, ISO_OFFSET_DATE_TIME, ISO_LOCAL_DATE_TIME) before reaching the formatter that actually matches. The old (buggy) single-formatter approach paid this cost once per call via internal optional-section backtracking, not via N discrete thrown/caught exceptions.
Fix: consider a cheap upfront discriminator (e.g. presence of 'T'/'Z' or a space at index 10) to pick the right formatter directly instead of trial-and-error, or at minimum reorder TIMESTAMP_FORMATTERS so the JDBC/space format — described here as the common case — is tried before the offset/zoned ISO variants that are comparatively rare for this driver-facing code path.

plugin-jdbc/src/test/java/io/kestra/plugin/jdbc/AbstractCellConverterTest.java
```diff
@@ -94,6 +94,23 @@ void testStringValueToTimestamp_JdbcFormat() throws Exception {
         );
     }
 
+    @Test
+    void testStringValueToTimestamp_JdbcFormatWithFractionalSeconds() throws Exception {
```
fdelbrayelle, line 101: 🟡 [Tests] No coverage for the new no-match failure path
Problem: good happy-path regression coverage for .1 / .123 / .123456789, but this diff also changes what happens when no formatter matches at all — previously an uncaught DateTimeParseException could propagate out of formatter.parseBest, now the loop exhausts and throws a clean IllegalArgumentException("Unsupported date format: " + value). That new terminal throw has no test exercising it.
Fix: add a case with an unparseable string (e.g. a >9-digit fraction, or garbage text) asserting the failure surfaces as the wrapped, actionable message from addPreparedStatementException (per the Error Messages guideline). Also worth adding the .12 (2-digit fraction) case explicitly since it's called out in issue #991's repro list.

fdelbrayelle (approved):
## Kestra Plugin Code Review

The generic 1-9 digit fractional-seconds formatter correctly covers all repro cases from #991 (.1, .12, .123, .123456789), matching `java.sql.Timestamp.toString()` output. Root-cause analysis (sequential vs. alternative `appendOptional` sections) is accurate, and the fix (independent formatters tried in a loop) is a correct, idiomatic solution to that root cause.

### Kestra Guidelines — 0 findings
This is a shared internal utility method (`AbstractCellConverter`), not a Task/Trigger, so most annotation/property/doc conventions don't apply. Change is scoped correctly to the shared base class per the stated root cause (single fix point for all 23 DB modules). The added Javadoc on `TIMESTAMP_FORMATTERS` documents a genuinely non-obvious WHY (why independent attempts are required instead of chained `appendOptional`), consistent with the comment guidelines. `TIMESTAMP_FORMATTERS` as a `static final List.of(...)` is a good immutability/perf improvement over the old per-call `DateTimeFormatterBuilder` construction. No violations found.

### Security (OWASP Top 10:2025 + KPS) — 0 issues found
No secrets, no external calls, no injection surface. Input is caller-controlled data already reaching this method pre-fix; parsing failure now fails closed with a clean `IllegalArgumentException` instead of an uncaught `DateTimeParseException`, which is a minor improvement, not a regression.

### Performance — 1 finding
🟡 Suggestion — Exception-driven sequential formatter matching on a hot path. See inline comment.

### Tests — 1 finding
🟡 Suggestion — No test for the new no-match failure path. See inline comment.

**Verdict:** APPROVE

karthikchundi-commits:
Thanks for the review @fdelbrayelle. Both suggestions addressed in 957060b9:

**Performance — exception-driven matching on the hot path**
`parseStringToTimestamp` now does a cheap separator check first (`charAt(10)`): a space → JDBC formatter, `T`/`t` → ISO local date-time, length 10 → bare date. A well-formed value is parsed on the first attempt with no failed-parse exceptions. The full candidate list is still tried as a fallback for the rarer offset/zoned ISO shapes, so behaviour is unchanged. `TIMESTAMP_FORMATTERS` is also reordered with the JDBC/space format first.

**Tests — no coverage for the no-match failure path**
Added `testStringValueToTimestamp_UnparseableSurfacesActionableError`: an unparseable value (`2019-10-30 12:00:00.1234567890`, >9 fraction digits) now asserts the failure surfaces as the wrapped `Exception` from `addPreparedStatementException` (message contains the value; cause is the `IllegalArgumentException` "Unsupported date format"), not a raw `DateTimeParseException`. Also added the 2-digit fraction (`.12`) case from #991's repro list.

`./gradlew :plugin-jdbc:test` — 43/43 passing locally (JDK 21). Rebased onto the latest `main` merge.

PR #996, opened 2026-08-31
fix(jdbc): duckdb.Query with afterSQL always fails with Connection was closed

### LINKED ISSUE #988: [Bash2.0] duckdb.Query with afterSQL always fails with Connection was closed
### Describe the issue

`duckdb.Query` with `afterSQL` always fails with `Connection was closed`. The connection is closed before `afterSQL` runs, and the rollback attempt then throws.

`duckdb.Queries` with the identical `afterSQL` succeeds, and `sqlite.Query` with the identical `afterSQL` succeeds, so it is specific to `duckdb.Query`. It fails on every `fetchType`, and removing `afterSQL` makes the same task pass.

```
Executing afterSQL: SELECT 1
java.sql.SQLException: Connection was closed
  at org.duckdb.DuckDBConnection.checkOpen(DuckDBConnection.java:583)
  at io.kestra.plugin.jdbc.AbstractJdbcQuery.rollbackIfTransactional(AbstractJdbcQuery.java:212)
```

### Visuals and Reproducibility

Image

Run the flow. The first and third tasks succeed, the second fails.

```yaml
id: v_duckdb_after_sql
namespace: company.team

tasks:
  - id: without_after_sql
    fetchType: FETCH_ONE
    sql: SELECT 42 AS answer

  - id: with_after_sql
    fetchType: FETCH_ONE
    sql: SELECT 42 AS answer
    afterSQL: SELECT 1

  - id: queries_with_after_sql
    type: io.kestra.plugin.jdbc.duckdb.Queries
    fetchType: FETCH
    sql: SELECT 42 AS answer
    afterSQL: SELECT 1
```

### Environment

Kestra Version: 2.0.0-rc10, commit 9900dcf
Setup: Standalone with Postgres, Kestra EE

nancysangani:
### Summary

Fixes #988 where `duckdb.Query` with `afterSQL` always failed due to two bugs:

* **Double `execute()`** in `executeAfterSQL()` caused DuckDB to execute the same prepared statement twice.
* **Connection closed too early** in `run()` meant rollback happened on a closed connection, masking the real error with `Connection was closed`.

### Fix

* Execute `afterSQL` only once when parameters are `null`.
* Explicitly close the connection in `finally`, **after** rollback in `catch`.

### Tests added

* `afterSQL` works without parameters.
* `afterSQL` runs in the same transaction.
* Actual `afterSQL` errors propagate instead of being replaced by `Connection was closed`.

### DIFF
--- a/plugin-jdbc/src/main/java/io/kestra/plugin/jdbc/AbstractJdbcQuery.java
+++ b/plugin-jdbc/src/main/java/io/kestra/plugin/jdbc/AbstractJdbcQuery.java
@@ -50,7 +50,7 @@ public abstract class AbstractJdbcQuery extends AbstractJdbcBaseQuery implements
         // on failure, the rollback attempt below runs against a still-open connection. With try-with-resources,
         // the resource is closed as the try block exits, which happens BEFORE the catch clause runs, causing
         // strict drivers (e.g. DuckDB) to fail the rollback with "Connection was closed" and mask the real error.
-        Connection conn = this.connection(runContext);
+        var conn = this.connection(runContext);
         try {
             this.runningConnection = conn;
             this.beforeExecute(runContext, conn);
@@ -141,16 +141,6 @@ public abstract class AbstractJdbcQuery extends AbstractJdbcBaseQuery implements
         } finally {
             this.runningConnection = null;
             closeConnection(runContext, conn);
-        }
-    }
-
-    private static void closeConnection(final RunContext runContext, final Connection connection) {
-        try {
-            if (connection != null) {
-                connection.close();
-            }
-        } catch (SQLException e) {
-            runContext.logger().warn("Issue when closing the connection : {}", e.getMessage());
         }
     }
 
--- a/plugin-jdbc/src/main/java/io/kestra/plugin/jdbc/AbstractJdbcQuery.java
+++ b/plugin-jdbc/src/main/java/io/kestra/plugin/jdbc/AbstractJdbcQuery.java
@@ -46,7 +46,12 @@ public abstract class AbstractJdbcQuery extends AbstractJdbcBaseQuery implements
         Savepoint savepoint = null;
         boolean supportsTx = false;
 
-        try (Connection conn = this.connection(runContext)) {
+        // Connection is closed explicitly in the finally block (instead of via try-with-resources) so that,
+        // on failure, the rollback attempt below runs against a still-open connection. With try-with-resources,
+        // the resource is closed as the try block exits, which happens BEFORE the catch clause runs, causing
+        // strict drivers (e.g. DuckDB) to fail the rollback with "Connection was closed" and mask the real error.
+        Connection conn = this.connection(runContext);
+        try {
             this.runningConnection = conn;
             this.beforeExecute(runContext, conn);
 
@@ -135,6 +140,17 @@ public abstract class AbstractJdbcQuery extends AbstractJdbcBaseQuery implements
             throw e;
         } finally {
             this.runningConnection = null;
+            closeConnection(runContext, conn);
+        }
+    }
+
+    private static void closeConnection(final RunContext runContext, final Connection connection) {
+        try {
+            if (connection != null) {
+                connection.close();
+            }
+        } catch (SQLException e) {
+            runContext.logger().warn("Issue when closing the connection : {}", e.getMessage());
         }
     }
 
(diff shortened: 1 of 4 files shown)


plugin-jdbc/src/main/java/io/kestra/plugin/jdbc/AbstractJdbcQuery.java
```diff
@@ -135,6 +140,17 @@ public AbstractJdbcBaseQuery.Output run(RunContext runContext) throws Exception
             throw e;
         } finally {
             this.runningConnection = null;
+            closeConnection(runContext, conn);
+        }
+    }
+
+    private static void closeConnection(final RunContext runContext, final Connection connection) {
```
fdelbrayelle: 🟠 [Guidelines] DRY violation — duplicated connection-close helper
Problem: `closeConnection` here is byte-for-byte identical (including the log message "Issue when closing the connection : {}") to `AbstractJdbcQueries.safelyCloseConnection()`, a sibling class that shares the common superclass `AbstractJdbcBaseQuery`. That base class already centralizes connection-lifecycle helpers (`kill(Connection)`), so this is the natural home for a shared close helper instead of two private copies under different names.
Fix: Move this method (or delegate to it) onto `AbstractJdbcBaseQuery` as a protected/static helper, and have both `AbstractJdbcQuery` and `AbstractJdbcQueries` call the single shared implementation.

plugin-jdbc/src/main/java/io/kestra/plugin/jdbc/AbstractJdbcQuery.java
```diff
@@ -46,7 +46,12 @@ public AbstractJdbcBaseQuery.Output run(RunContext runContext) throws Exception
         Savepoint savepoint = null;
         boolean supportsTx = false;
 
-        try (Connection conn = this.connection(runContext)) {
+        // Connection is closed explicitly in the finally block (instead of via try-with-resources) so that,
+        // on failure, the rollback attempt below runs against a still-open connection. With try-with-resources,
+        // the resource is closed as the try block exits, which happens BEFORE the catch clause runs, causing
+        // strict drivers (e.g. DuckDB) to fail the rollback with "Connection was closed" and mask the real error.
+        Connection conn = this.connection(runContext);
```
fdelbrayelle: 🟢 [Guidelines] Use var for the new local variable
Problem: `Connection conn = this.connection(runContext);` is a newly introduced local declaration with an unambiguous right-hand-side type.
Fix: `var conn = this.connection(runContext);` per the repo's "use var for all local variables whenever legally possible" convention.

fdelbrayelle (approved):
## Kestra Plugin Code Review

The PR fixes the two bugs described in #988: the double `execute()` call on the `afterSQL` `PreparedStatement`, and the try-with-resources ordering that closed the connection before the `catch` block's rollback ran. Tests cover the happy path (no `afterSQL` failure), the transactional effect of `afterSQL`, and that a genuine `afterSQL` SQL error now propagates instead of being masked. Scope is appropriate — no scope creep.

### Kestra Guidelines — 2 findings
- 🟠 DRY violation: `AbstractJdbcQuery.closeConnection()` duplicates `AbstractJdbcQueries.safelyCloseConnection()` line-for-line (identical try/catch, identical log message), even though both classes already share a common superclass, `AbstractJdbcBaseQuery`, which already hosts equivalent shared connection-lifecycle helpers (`kill(Connection)`). This should be a single protected/static helper on `AbstractJdbcBaseQuery` instead of two private copies with different names.
- 🟢 Nit: `Connection conn = this.connection(runContext);` (line 53) is a new local variable declaration with an unambiguous inferable type — per the repo's Java style guideline, this should be `var conn = ...`.

### Security (OWASP Top 10:2025 + KPS) — no issues found
No new SQL injection surface (rendering of `afterSQL` is unchanged and still goes through `runContext.render`). No credential handling touched. The new `closeConnection` catch-and-log does not leak connection strings/credentials in the log message (`e.getMessage()` on a close failure is low-risk here and matches the existing pattern used in `AbstractJdbcQueries`).

### Performance — no issues found
No unbounded allocation, no new blocking loops, no new shared mutable state. Resource handling (statements, result sets) unchanged aside from the connection lifecycle fix itself.

### Correctness note (non-blocking)
Reordering `this.runningConnection = null;` before `closeConnection(runContext, conn)` in the `finally` block (previously the connection was physically closed before the field was nulled, since try-with-resources closed it ahead of the outer `finally`) introduces a small window where a concurrent `kill()` call would see `runningConnection == null` and no-op while the connection is still technically open at that instant. This is a negligible race (the window is a single statement) and not worse than the pre-existing behavior in practice, but closing the connection before nulling the field would be strictly safer and preserve the original ordering guarantee for `kill()`.

**Verdict:** APPROVE

plugin-jdbc/src/main/java/io/kestra/plugin/jdbc/AbstractJdbcQuery.java
nancysangani: fixed — moved `closeConnection` up to `AbstractJdbcBaseQuery` as a shared protected static helper and had both `AbstractJdbcQuery` and `AbstractJdbcQueries` call it, removing the old `safelyCloseConnection` duplicate.

plugin-jdbc/src/main/java/io/kestra/plugin/jdbc/AbstractJdbcQuery.java
nancysangani: Done - switched to `var conn = this.connection(runContext);`

PR #997, opened 2026-09-01
fix(tests): wait for containers with a loop instead of a task-level retry

jymaire:
## Problem

The nightly `main.yml` run on `main` was red on the Trino module ([run 33479298525](https://github.com/kestra-io/plugin-jdbc/actions/runs/33479298525)) — the only failure in the suite (`388 total · 334 passed · 1 failed · 53 skipped`):

```
java.lang.AssertionError:
Expected: a collection with size <10>
     but: collection size was <11>
    at io.kestra.plugin.jdbc.trino.RunnerTest.all_trino(RunnerTest.java:19)
```

The execution itself was fine — the same CI log shows `all_trino` … `Flow completed with state SUCCESS in 00:00:20.543`. Only the hard-coded task-run count failed.

## Root cause

`all_trino.yaml` declared a task-level `retry` on `wait_for_trino` so it could wait for the Trino container to accept connections. When that retry fires, the 11th entry is **a second, distinct task run for the same `wait_for_trino` task id** — not an extra attempt on the existing one. The CI log shows both ids under one execution (`50yDkGhBDXDxJ6v2JbXyJ0`):

```
[task: wait_for_trino] [taskrun: nKt4jVahJ2DYSjn5iWpFP]  -> FAILED, kept retrying
[task: wait_for_trino] [taskrun: 2ZKlEd5lYH4Jc8JMBVkTKg] -> SUCCESS, flow proceeded
```

`ExecutorService` marks the failed task run `RETRYING` and schedules a `RESTART_FAILED_TASK` delay; `ExecutionService.retryTask` then reuses the *same* task run id and `Execution.withTaskRun` replaces the entry in place — so a retry should never append. The duplicate appears around the `RETRYING` transition (the second run starts ~250 ms after the first failure, well inside the retry interval), which makes it **timing-dependent**.

Bumping the expected number to `11` would therefore have been wrong: the extra entry is a race, not a constant.

### The same race can silently skip a task

While investigating, ~2 of 6 local Trino runs and **every** local ClickHouse run showed the duplicate **dropping the last task of the `tasks:` list** — `assert` never executed — while the execution still reported `SUCCESS`.

That means the old assertions could go **falsely green**. ClickHouse produced 12 distinct tasks (no `assert`) + 1 duplicate `wait_for_clickhouse` task run = 13 raw entries, exactly what `hasSize(13)` demanded. The sanity check passed without ever running its only validation step.

This is a real defect in the Kestra executor's retry path. Filed upstream as kestra-io/kestra#18909, and referenced from the rationale comment in each flow.

## Change

Replace the task-level `retry` on every readiness task with a `LoopUntil` around an `allowFailure` probe:

```yaml
- id: wait_for_trino
  type: io.kestra.plugin.core.flow.LoopUntil
  condition: "{{ (outputs.trino_ping ?? {}).row is defined }}"
  failOnMaxReached: true
  checkFrequency:
    interval: PT5S
    maxDuration: PT300S
  tasks:
    - id: trino_ping
      type: io.kestra.plugin.jdbc.trino.Query
      username: "{{ vars.trino_user }}"
      url: "jdbc:trino://localhost:{{ outputs.port.value }}"
      sql: "SELECT 1"
      fetchType: FETCH_ONE
      allowFailure: true
```

A refused connection just ends the iteration instead of failing the task, so no retry is triggered and no task run is duplicated. The loop exits as soon as the probe returns a row, and `failOnMaxReached: true` makes it fail loudly if the container never becomes reachable.

`LoopUntil` creates a **new** `TaskRun` for its child on every polling iteration, so the readiness wait contributes `1 + N` task runs, where `N` depends on how long the container took to come up. The assertions therefore move off the raw `getTaskRunList()` size and onto the number of **distinct task ids**:

```java
assertThat(execution.getTaskRunList().stream().map(TaskRun::getTaskId).distinct().toList(), hasSize(N));
```

This is immune to the iteration count while still catching the skipped-task failure mode above — a task that never runs shrinks the distinct count. A looser matcher such as `greaterThanOrEqualTo(...)` was deliberately avoided: it would have kept hiding that bug.

The `SUCCESS` state assertion is unchanged.

### Modules changed (one commit each)

| Module | Expected distinct task ids |
| --- | --- |
| `plugin-jdbc-trino` | 11 |
| `plugin-jdbc-clickhouse` | 14 |
| `plugin-jdbc-db2` | 12 |
| `plugin-jdbc-oracle` | 13 |
| `plugin-jdbc-sybase` | 9 |

Outer `@ExecuteFlow` timeouts are also raised where the loop's `checkFrequency.maxDuration` would otherwise outlive them, so JUnit cannot kill the test before `failOnMaxReached: true` fires:

| Module | Inner `maxDuration` | Outer `timeout` |
| --- | --- | --- |
| `plugin-jdbc-trino` | `PT300S` | `PT600S` |
| `plugin-jdbc-clickhouse` | `PT300S` | `PT600S` |
| `plugin-jdbc-db2` | `PT900S` | `PT1500S` |
| `plugin-jdbc-sybase` | `PT900S` | `PT1500S` |
| `plugin-jdbc-oracle` | `PT1200S` | `PT1800S` |

### Modules deliberately left alone

`plugin-jdbc-access`, `plugin-jdbc-duckdb`, `plugin-jdbc-mariadb`, `plugin-jdbc-mysql`, `plugin-jdbc-postgres`, `plugin-jdbc-sqlite`, `plugin-jdbc-sqlserver` — no `retry` in the flow, so they are not exposed to this race.

## Verification

Run locally with Docker and the pinned `kestraVersion=2.0.0-rc12` (same as CI).

- Reproduced the CI failure exactly on `main`: `collection size was <11>`, `assert` ran, execution `SUCCESS`.
- With this change, `:plugin-jdbc-trino:test --tests '*RunnerTest*'` passes 3/3 consecutive runs.
- `:plugin-jdbc-clickhouse:test --tests '*RunnerTest*'` passes, and `assert` now actually executes on every run.
- `:plugin-jdbc-db2`, `:plugin-jdbc-oracle`, `:plugin-jdbc-sybase` test sources compile — their tests are `@Disabled` for runner disk space, so they cannot be executed here.
- `spotlessCheck` passes.
- CI on this branch: the modules touched by this PR pass. `plugin-jdbc-mariadb` is currently failing (`Expected size <11> but was <6>`, `socket was closed by server`) — that module is **untouched** by this PR and has no `retry` in its flow; it looks like the same container-readiness flakiness and is tracked separately.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/plugin-jdbc-clickhouse/src/test/java/io/kestra/plugin/jdbc/clickhouse/RunnerTest.java
+++ b/plugin-jdbc-clickhouse/src/test/java/io/kestra/plugin/jdbc/clickhouse/RunnerTest.java
@@ -3,6 +3,7 @@ package io.kestra.plugin.jdbc.clickhouse;
 import io.kestra.core.junit.annotations.ExecuteFlow;
 import io.kestra.core.junit.annotations.KestraTest;
 import io.kestra.core.models.executions.Execution;
+import io.kestra.core.models.executions.TaskRun;
 import io.kestra.core.models.flows.State;
 import org.junit.jupiter.api.Test;
 
@@ -16,7 +17,9 @@ class RunnerTest {
     @Test
     @ExecuteFlow(value = "sanity-checks/all_clickhouse.yaml", timeout = "PT600S")
     void all_clickhouse(Execution execution) {
-        assertThat(execution.getTaskRunList(), hasSize(13));
+        // the readiness task retries until the container is up, which can duplicate its task run,
+        // so assert on distinct task ids rather than the raw task run count
+        assertThat(execution.getTaskRunList().stream().map(TaskRun::getTaskId).distinct().toList(), hasSize(13));
         assertThat(execution.getState().getCurrent(), is(State.Type.SUCCESS));
     }
 }
--- a/plugin-jdbc-clickhouse/src/test/java/io/kestra/plugin/jdbc/clickhouse/RunnerTest.java
+++ b/plugin-jdbc-clickhouse/src/test/java/io/kestra/plugin/jdbc/clickhouse/RunnerTest.java
@@ -3,7 +3,6 @@ package io.kestra.plugin.jdbc.clickhouse;
 import io.kestra.core.junit.annotations.ExecuteFlow;
 import io.kestra.core.junit.annotations.KestraTest;
 import io.kestra.core.models.executions.Execution;
-import io.kestra.core.models.executions.TaskRun;
 import io.kestra.core.models.flows.State;
 import org.junit.jupiter.api.Test;
 
@@ -17,9 +16,7 @@ class RunnerTest {
     @Test
     @ExecuteFlow(value = "sanity-checks/all_clickhouse.yaml", timeout = "PT600S")
     void all_clickhouse(Execution execution) {
-        // the readiness task retries until the container is up, which can duplicate its task run,
-        // so assert on distinct task ids rather than the raw task run count
-        assertThat(execution.getTaskRunList().stream().map(TaskRun::getTaskId).distinct().toList(), hasSize(13));
+        assertThat(execution.getTaskRunList(), hasSize(14));
         assertThat(execution.getState().getCurrent(), is(State.Type.SUCCESS));
     }
 }
--- a/plugin-jdbc-clickhouse/src/test/java/io/kestra/plugin/jdbc/clickhouse/RunnerTest.java
+++ b/plugin-jdbc-clickhouse/src/test/java/io/kestra/plugin/jdbc/clickhouse/RunnerTest.java
@@ -3,6 +3,7 @@ package io.kestra.plugin.jdbc.clickhouse;
 import io.kestra.core.junit.annotations.ExecuteFlow;
 import io.kestra.core.junit.annotations.KestraTest;
 import io.kestra.core.models.executions.Execution;
+import io.kestra.core.models.executions.TaskRun;
 import io.kestra.core.models.flows.State;
 import org.junit.jupiter.api.Test;
 
@@ -16,7 +17,7 @@ class RunnerTest {
     @Test
     @ExecuteFlow(value = "sanity-checks/all_clickhouse.yaml", timeout = "PT600S")
     void all_clickhouse(Execution execution) {
-        assertThat(execution.getTaskRunList(), hasSize(14));
+        assertThat(execution.getTaskRunList().stream().map(TaskRun::getTaskId).distinct().toList(), hasSize(14));
         assertThat(execution.getState().getCurrent(), is(State.Type.SUCCESS));
     }
 }
--- a/plugin-jdbc-clickhouse/src/test/resources/sanity-checks/all_clickhouse.yaml
+++ b/plugin-jdbc-clickhouse/src/test/resources/sanity-checks/all_clickhouse.yaml
@@ -21,7 +21,7 @@ tasks:
       CLICKHOUSE_PASSWORD: "{{ vars.clickhouse_password }}"
 
   # Poll for readiness with a loop rather than a task-level `retry`: a retry makes the executor
-  # duplicate the task run, which can silently drop a downstream task (see kestra-io/kestra).
+  # duplicate the task run, which can silently drop a downstream task (see kestra-io/kestra#18909).
   # The probe is `allowFailure` so a refused connection just ends the iteration; the loop exits
   # once the probe returned a row, and fails loudly if the container never becomes reachable.
   - id: wait_for_clickhouse
--- a/plugin-jdbc-clickhouse/src/test/resources/sanity-checks/all_clickhouse.yaml
+++ b/plugin-jdbc-clickhouse/src/test/resources/sanity-checks/all_clickhouse.yaml
@@ -20,17 +20,26 @@ tasks:
     env:
       CLICKHOUSE_PASSWORD: "{{ vars.clickhouse_password }}"
 
+  # Poll for readiness with a loop rather than a task-level `retry`: a retry makes the executor
+  # duplicate the task run, which can silently drop a downstream task (see kestra-io/kestra).
+  # The probe is `allowFailure` so a refused connection just ends the iteration; the loop exits
+  # once the probe returned a row, and fails loudly if the container never becomes reachable.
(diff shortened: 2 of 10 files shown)


fdelbrayelle (changes requested):
Thanks for the thorough root-cause on the `retry` duplication — dropping the task-level `retry` in favor of `LoopUntil` + `allowFailure` is the right call for that race. A few things before merge:

### 🔴 Blocking: exact `hasSize(N)` assertions are still non-deterministic under `LoopUntil`

`LoopUntil` does not reuse a single task run for its child across polling iterations. Each iteration creates a new `TaskRun` for the probe task (`trino_ping`, `db2_ping`, …) and appends it to `execution.getTaskRunList()`. See `ResolvedTask.toNextTaskRunIncrementIteration` in kestra core (calls `TaskRun.of(...)` per iteration) and core's own `LoopUntilCaseTest.waitforNestedThreeLevels`, which asserts multiple distinct task runs sharing the same task id across iterations.

So the extra count contributed by `wait_for_<db>` is `1 (LoopUntil) + N (polling iterations)`, not a constant `+1` as the PR description states. `N` depends on container start time — and for db2 / oracle / sybase, needing several polls is the *normal* path (that's why `maxDuration` is 900–1200s). `hasSize(12)` / `hasSize(13)` / `hasSize(9)` will fail on any run where the DB doesn't answer on the very first poll.

**Suggested fix:** keep the `LoopUntil` change, but assert on distinct task ids instead of raw task-run count (the approach from the reverted intermediate commit `10739f54`):

```java
assertThat(
    execution.getTaskRunList().stream().map(TaskRun::getTaskId).distinct().toList(),
    hasSize(N)
);
```

This still catches the "assert task silently skipped" failure mode (missing task id shrinks the distinct count) while being immune to however many iterations the loop needed. Applies to all 5 `RunnerTest.java` files touched here.

### 🟠 db2 / sybase: `checkFrequency.maxDuration` exceeds the outer `@ExecuteFlow` timeout

- `all_db2.yaml` → `maxDuration: PT900S`, but `db2/RunnerTest.java` still has `timeout = "PT600S"`
- `all_sybase.yaml` → `maxDuration: PT900S`, but `sybase/RunnerTest.java` still has `timeout = "PT600S"`

JUnit will kill the test before the loop's `failOnMaxReached: true` ever fires for any container that comes up between 600s and 900s. Oracle got this right in the same PR (`PT1800S` outer vs `PT1200S` inner) — db2 and sybase were missed. Currently masked because both tests are `@Disabled`, but worth fixing now so they don't ship broken when re-enabled.

### 🟡 Nit: duplicated rationale comment with dangling reference

The same 4-line comment is copy-pasted into all 5 YAML files and references "kestra-io/kestra" without an issue number. Could you file the upstream issue for the retry-duplication bug and link it directly (`see kestra-io/kestra#<n>`)?

### FYI: current CI on this branch is red

Latest `check / main` run (`33602462551`) fails in `plugin-jdbc-mariadb` (`Expected: a collection with size <11> but: collection size was <6>`, `socket was closed by server`). That module is untouched by this PR and the failure looks like unrelated container-readiness flakiness — not blocking — but the "CI is green" note in the description is stale relative to the branch tip.

---
Security / performance: no concerns (test-only diff, no new deps, no credential changes).

fdelbrayelle (approved):

jymaire:
Thanks for the review — all three points addressed.

### 🔴 `hasSize(N)` non-determinism — fixed in 1fe072c1

You're right, and the PR description was wrong to claim a constant `+1`. `LoopUntil` allocates a new `TaskRun` per iteration, so the readiness wait contributes `1 + N`, and for db2/oracle/sybase several polls are the normal path.

All 5 `RunnerTest.java` files now assert on distinct task ids, as you suggested:

```java
assertThat(execution.getTaskRunList().stream().map(TaskRun::getTaskId).distinct().toList(), hasSize(N));
```

The `N` values didn't need to change — the counts already in the diff happen to be exactly the distinct-id counts. I checked each flow's task ids to confirm, and none of the five contain any other multi-run construct (`ForEach`, `Parallel`, …) that would make raw count and distinct count diverge elsewhere:

| Module | Distinct task ids |
| --- | --- |
| `plugin-jdbc-trino` | 11 |
| `plugin-jdbc-clickhouse` | 14 |
| `plugin-jdbc-db2` | 12 |
| `plugin-jdbc-oracle` | 13 |
| `plugin-jdbc-sybase` | 9 |

### 🟠 db2 / sybase outer timeout — fixed in 1fe072c1

Good catch, those two were missed. Both raised `PT600S` → `PT1500S`, keeping Oracle's 1.5× headroom over the inner `maxDuration`:

| Module | Inner `maxDuration` | Outer `timeout` |
| --- | --- | --- |
| `plugin-jdbc-trino` | `PT300S` | `PT600S` |
| `plugin-jdbc-clickhouse` | `PT300S` | `PT600S` |
| `plugin-jdbc-db2` | `PT900S` | `PT1500S` |
| `plugin-jdbc-sybase` | `PT900S` | `PT1500S` |
| `plugin-jdbc-oracle` | `PT1200S` | `PT1800S` |

### 🟡 Dangling reference — fixed in b6d31c30

Filed the upstream bug as kestra-io/kestra#18909 (repro flow, the duplicate task-run ids from the CI log, the silently-skipped-`assert` failure mode, and the `LoopUntil` workaround). The comment in all 5 flows now reads `(see kestra-io/kestra#18909)`.

I left the comment duplicated across the 5 YAMLs rather than factoring it out — they're independent sanity-check flows with no shared include mechanism — but happy to trim it to a one-liner + issue link if you'd prefer less repetition.

### CI note

Updated the description: the "CI is green" line was stale, and it also still described the old `+1` / raw-`getTaskRunList()` approach. It now reflects the distinct-id assertions and states the actual CI state, including that `plugin-jdbc-mariadb` is red and untouched by this PR.

Agreed the mariadb failure is unrelated — that flow has no `retry`, so it isn't exposed to this race; it looks like plain container-readiness flakiness. Should I fold a `LoopUntil` readiness probe for mariadb into this PR, or keep this one scoped to the 5 `retry` flows and handle mariadb separately?

fdelbrayelle (approved):

PR #998, opened 2026-09-01
fix: bump bundled DuckDB ion extension to v1.5.5

### LINKED ISSUE #989: [Bash2.0] Bundled DuckDB ion extension is built for v1.5.3 but the driver is v1.5.5
### Describe the issue

Every DuckDB task run logs a warning because the bundled `ion` extension is built for a different DuckDB version than the driver.

```
Failed to load bundled DuckDB extension 'ion'. Falling back to community install.
message=Invalid Input Error: Failed to load '.../linux_arm64/ion.duckdb_extension',
The file was built specifically for DuckDB version 'v1.5.3' and can only be loaded with that version of DuckDB.
(this version of DuckDB is 'v1.5.5')
```

The community install fallback works, so nothing is broken functionally, but it means an outbound download on a plugin that otherwise needs no external dependency. That will fail in an air-gapped install. `communityExtensions` defaults to `["ion"]`, so it cannot be avoided without overriding the default.

Reproduces on any DuckDB task, including `sql: SELECT 1`.

### Visuals and Reproducibility

Image

```yaml
id: duckdb_ion_extension_version
namespace: company.team

tasks:
  - id: trivial
    fetchType: FETCH_ONE
    sql: SELECT 1
```

### Environment

Kestra Version: 2.0.0-rc10, commit 9900dcf
Setup: Standalone with Postgres, Kestra EE

nancysangani:
Fixes #989. `plugin-jdbc-duckdb/build.gradle` pinned `duckdbCommunityExtensionVersion` to `v1.5.3` while the driver dependency (`duckdb_jdbc:1.5.5.1`) ships DuckDB core v1.5.5, so the bundled ion extension was rejected at load time (`built for v1.5.3 ... this version of DuckDB is v1.5.5`), silently falling back to a network install of the community extension.

Fix: bump `duckdbCommunityExtensionVersion` to `v1.5.5` so the bundled binary matches the driver's actual DuckDB core version.

Verified:
- `https://community-extensions.duckdb.org/v1.5.5/{linux_amd64,linux_arm64}/ion.duckdb_extension.gz` both resolve (200).
- Confirmed `duckdb_jdbc:1.5.5.1`'s actual core version via `DatabaseMetaData.getDatabaseProductVersion()` → `v1.5.5`.
- `./gradlew :plugin-jdbc-duckdb:bundleDuckDbCommunityExtensions` re-downloads and bundles the v1.5.5 ion extension successfully.

### DIFF
--- a/plugin-jdbc-duckdb/build.gradle
+++ b/plugin-jdbc-duckdb/build.gradle
@@ -1,7 +1,7 @@
 project.description = 'Manage data in DuckDB with Kestra\'s JDBC plugin.'
 
 // Keep duckdb_jdbc inline for Dependabot; update duckdbCommunityExtensionVersion in sync.
-def duckdbCommunityExtensionVersion = "v1.5.3"
+def duckdbCommunityExtensionVersion = "v1.5.5"
 def bundledDuckDbExtensionsDir = layout.buildDirectory.dir("generated/duckdb-community-extensions")
 def bundledDuckDbExtensions = [
     ion: ["linux_amd64", "linux_arm64"]


plugin-jdbc-duckdb/build.gradle
```diff
@@ -1,7 +1,7 @@
 project.description = 'Manage data in DuckDB with Kestra\'s JDBC plugin.'
 
 // Keep duckdb_jdbc inline for Dependabot; update duckdbCommunityExtensionVersion in sync.
-def duckdbCommunityExtensionVersion = "v1.5.3"
+def duckdbCommunityExtensionVersion = "v1.5.5"
```
fdelbrayelle, line 4: 🟡 [Guidelines] Add a regression test for bundled-extension load success
Problem: DEFAULT_COMMUNITY_EXTENSIONS already includes "ion", so existing tests already exercise tryLoadBundledExtension(), but the fallback-to-community-install path only logs a WARN and is swallowed — nothing asserts the bundled binary actually loaded. This exact version drift (v1.5.3 pin vs v1.5.5 driver) shipped silently for that reason.
Fix: Add a test that queries duckdb_extensions() after a query run and asserts installed=true/loaded=true with install_path pointing at the extracted bundled file (or captures logs and asserts no "Failed to load bundled DuckDB extension" WARN is emitted), so a future duckdb_jdbc bump without a matching duckdbCommunityExtensionVersion bump fails CI instead of silently falling back.

fdelbrayelle (approved):
## Kestra Plugin Code Review

The PR fixes exactly the reported root cause: `duckdbCommunityExtensionVersion` (v1.5.3) was out of sync with the DuckDB core version shipped by `duckdb_jdbc:1.5.5.1` (v1.5.5), causing the bundled `ion` extension to be rejected at load time and silently falling back to a network install — breaking air-gapped installs as described in #989. Bumping the pin to v1.5.5 aligns the bundled binary with the driver's actual core version. Verified the version is read from a single generated metadata source (`DuckDbConnectionSetup.bundledDuckDbVersion()` reads `duckdb.version` from the properties file generated by this same Gradle variable), so there is no other hardcoded copy of the version left stale.

### Kestra Guidelines — 1 finding
- 🟡 No regression test guards this exact defect class. `DuckDbQueryInterface.DEFAULT_COMMUNITY_EXTENSIONS` already includes `ion`, so every existing DuckDB test already exercises `tryLoadBundledExtension`, but the fallback path only logs a WARN and is swallowed (caught `SQLException`/`IOException`) — no test asserts the bundled extension actually loaded (e.g. querying `duckdb_extensions()` for `installed=true`/`loaded=true` with an `install_path` pointing at the extracted bundled file, or asserting no "Failed to load bundled DuckDB extension" WARN is logged). That's precisely how this version drift shipped unnoticed. Given the existing `// Keep duckdb_jdbc inline for Dependabot; update duckdbCommunityExtensionVersion in sync.` comment already flags the sync risk, a cheap assertion here would catch the next driver bump before it regresses.

### Security (OWASP Top 10:2025 + KPS) — 0 findings introduced by this diff
No new dependency, no secret, no HTTP client change in the reviewed hunk. Note (pre-existing, not introduced here, purely FYI): `bundleDuckDbCommunityExtensions` downloads the extension `.gz` over plain `URL#withInputStream` with no checksum/signature verification before it's unpacked and bundled into the shipped JAR (A03/A08 supply-chain surface). Not a blocker for this one-line version bump, but worth a follow-up if this task is touched again.

### Performance — 0 findings
Build-time only change; no runtime hot path affected.

**Verdict:** APPROVE

Issue #1002, opened 2026-09-04
Quoted identifiers unsupported: unquoted connection properties, Trigger dropping schema/role, and named-parameter substitution


jymaire:
Reported by a user whose Snowflake database is named `INGESTION:TOPIC_ENVIRONMENT`. A `snowflake.Trigger` on that database failed while an equivalent `snowflake.Query` task worked, which looked like the trigger not supporting quoted identifiers.

Investigating it turned up **three independent defects**. The first is the reported failure; the second contributes to it; the third produces the same symptom by a different route and affects every JDBC plugin.

---

### Expected

`database: INGESTION:TOPIC_ENVIRONMENT` should connect. Snowflake requires a name like this to be a quoted identifier, and the plugin should supply the quotes.

### Actual

The connection fails before any SQL is sent, because the value reaches the driver unquoted. `renderProperties` passes each identifier straight through:

https://github.com/kestra-io/plugin-jdbc/blob/1ec1018e8ce86934a8e17b184cf9a98a28236375/plugin-jdbc-snowflake/src/main/java/io/kestra/plugin/jdbc/snowflake/SnowflakeInterface.java#L125-L144

```java
if (this.getDatabase() != null) {
    var rDatabase = runContext.render(this.getDatabase()).as(String.class).orElseThrow();
    properties.put("db", rDatabase);
}
```

Same for `warehouse` (L128), `schema` (L138) and `role` (L143).

### User-confirmed workaround

Bypass the properties entirely and put the identifier in the JDBC URL, URL-encoded, with `%22` for the quotes:

```
jdbc:snowflake://<account>.snowflakecomputing.com/?role=yourrole&db=%22INGESTION%3ATOPIC_ENVIRONMENT%22
```

The URL query string and the `Properties` map feed the same driver settings via `DriverManager.getConnection(jdbcUrl, props)`, so the only thing this changes is that `db` arrives quoted. That is what confirms the diagnosis. Setting `database: '"INGESTION:TOPIC_ENVIRONMENT"'` (quotes inside the YAML value) should work equivalently.

### Suggested fix

Quote `db`, `schema`, `warehouse` and `role` when the value is not a legal unquoted Snowflake identifier (`[A-Za-z_][A-Za-z0-9_$]*`).

[...]

Issue #364, opened 2024-08-26
Snowflake trigger not working: choosing warehouse


MarthaScheffler:
### Expected Behavior

When I use the Snowflake jdbc **trigger**, I want to be able to choose the virtual warehouse used in that session.
Either via the Kestra parameter "warehouse", or by firing several SQL commands, "USE WAREHOUSE WH; SELECT 1;"
neither is working. only workaround is to give the authenticated user a default warehouse.

(the actual jdbc **tasks** also seem to work fine, at least for the Query I could set a warehouse and it got used)

### Actual Behaviour

Error when using the Kestra parameter "warehouse":
```
net.snowflake.client.jdbc.SnowflakeSQLException: No active warehouse selected in the current session. Select an active warehouse with the ‘use warehouse’ command.
```

Error when firing 2 SQL commands in the same session:
```
net.snowflake.client.jdbc.SnowflakeSQLException: Actual statement count 2 did not match the desired statement count 1.
```
multi-statement support from Snowflake's SQL-api: https://docs.snowflake.com/en/developer-guide/sql-api/submitting-multiple-statements#specifying-multiple-sql-statements-in-the-request

### Steps To Reproduce

1. using the code from https://kestra.io/plugins/plugin-jdbc-snowflake/triggers/io.kestra.plugin.jdbc.snowflake.trigger
2. setting additionally a warehouse
3. changing the sql to "SELECT WAREHOUSE my_warehouse; SELECT * FROM demo_db.public.customers;"

### Environment Information

- Kestra Version: 0.18.3
- Plugin version: ?
- Operating System (OS / Docker / Kubernetes): kubernetes
- Java Version (If not docker):

Ben8t:
@MarthaScheffler ping ;) this one is fixed !

PR #1003, opened 2026-09-04
fix(snowflake): support quoted identifiers in Trigger and connection properties

tchiotludo:

A `snowflake.Trigger` failed to connect to a Snowflake database whose name requires a quoted identifier (e.g. `INGESTION:TOPIC_ENVIRONMENT`, where the colon forces a quoted identifier), while an equivalent `snowflake.Query` worked. Investigating issue #1002 turned up two Snowflake-side defects addressed here:

1. **Identifier connection properties were never quoted.** `SnowflakeInterface.renderProperties()` injected `db`, `schema`, `warehouse` and `role` as raw connection properties, so a name that Snowflake can only accept as a quoted identifier reached the driver unquoted and the connection failed before any SQL was sent.

2. **`snowflake.Trigger` silently dropped `schema` and `role`.** The trigger does not run the query itself — it builds a `Query` and delegates to it — and that builder omitted `schema` and `role` (and the connection-pooling properties). A `Trigger` and a `Query` with identical YAML therefore connected with different sessions, with nothing indicating the values were discarded. This is the same class of defect as #364 (which forwarded only `warehouse` and `database`).

**Fixes made:**

- `Trigger.runQuery()` now propagates `schema`, `role`, `connectionPooling` and `connectionPoolSize` to the delegated `Query`, alongside the `warehouse`/`database` already forwarded.
- `SnowflakeInterface.renderProperties()` now passes `db`, `schema`, `warehouse` and `role` through a new conservative helper `quoteIdentifierIfNeeded(String)`:
  - a value that is already double-quoted is left unchanged;
  - a value matching the legal unquoted Snowflake identifier pattern `^[A-Za-z_][A-Za-z0-9_$]*$` is left unchanged, so ordinary names such as `INGESTION`/`mydb` keep today's case-normalizing behavior;
  - any other value (e.g. one containing a colon) is wrapped in double quotes, with embedded `"` escaped as `""`.

This intentionally does **not** quote unconditionally — quoting makes an identifier case-sensitive, so `database: mydb` (which resolves to `MYDB` today) must keep working.

**Tests added** (`SnowflakeInterfaceTest`, no live Snowflake needed):
- unit tests for `quoteIdentifierIfNeeded`: `INGESTION` stays unquoted, `INGESTION:TOPIC_ENVIRONMENT` becomes `"INGESTION:TOPIC_ENVIRONMENT"`, an already-quoted value is unchanged, embedded quotes are escaped;
- `renderProperties` tests asserting the resulting `Properties` map: colon-containing `db`/`role` are quoted while an ordinary `warehouse`/`schema` are not, and unset identifiers are omitted.

`./gradlew :plugin-jdbc-snowflake:compileJava :plugin-jdbc-snowflake:compileTestJava` and the added `SnowflakeInterfaceTest` (6 tests) pass. The live Snowflake integration tests remain `@Disabled` (no server / credentials), unchanged.

**Follow-up (not in this PR):** issue #1002 also describes a third, independent defect in the shared base class `AbstractJdbcBaseQuery` — when `parameters` is set, the `:\w+` named-parameter regex rewrites a colon inside a quoted identifier (and breaks the PostgreSQL `::` cast, string literals and comments). That fix requires a broader rewrite of the parameter tokenizer used by every JDBC plugin, so it is left for a maintainer to keep this change tight and low-risk. This PR uses `Refs #1002` rather than `Closes` for that reason.

---

Covered by the new unit tests. Illustrative flow that this fix unblocks:

```yaml
id: snowflake_quoted_db
namespace: company.team

triggers:
  - id: watch
    type: io.kestra.plugin.jdbc.snowflake.Trigger
    interval: "PT5M"
    url: jdbc:snowflake://<account_identifier>.snowflakecomputing.com
    username: "{{ secret('SNOWFLAKE_USERNAME') }}"
    password: "{{ secret('SNOWFLAKE_PASSWORD') }}"
    database: "INGESTION:TOPIC_ENVIRONMENT"
    schema: PUBLIC
    role: MY_ROLE
    warehouse: COMPUTE_WH
    sql: "SELECT * FROM customers"
    fetchType: FETCH
```

Before this change the `database` value reached the driver unquoted and the connection failed, and `schema`/`role` set on the trigger were ignored.

---

### Setup Instructions

No additional setup required.

---

### Contributor Checklist ✅

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01H616mqoRaPGwCzHYrKUHEm

---
_Generated by [Claude Code](https://claude.ai/code/session_01H616mqoRaPGwCzHYrKUHEm)_

### DIFF
--- a/plugin-jdbc-snowflake/src/main/java/io/kestra/plugin/jdbc/snowflake/SnowflakeInterface.java
+++ b/plugin-jdbc-snowflake/src/main/java/io/kestra/plugin/jdbc/snowflake/SnowflakeInterface.java
@@ -8,9 +8,16 @@ import io.kestra.plugin.jdbc.JdbcConnectionInterface;
 import io.swagger.v3.oas.annotations.media.Schema;
 
 import java.util.Properties;
+import java.util.regex.Pattern;
 
 
 public interface SnowflakeInterface extends JdbcConnectionInterface {
+
+    /**
+     * Matches a value that is a legal unquoted Snowflake identifier and therefore
+     * needs no quoting (case is normalized to upper-case by the server).
+     */
+    Pattern SAFE_UNQUOTED_IDENTIFIER = Pattern.compile("^[A-Za-z_][A-Za-z0-9_$]*$");
 
     @PluginProperty(group = "advanced")
     @Schema(
@@ -125,22 +132,22 @@ public interface SnowflakeInterface extends JdbcConnectionInterface {
     default void renderProperties(RunContext runContext, Properties properties) throws IllegalVariableEvaluationException {
         if (this.getWarehouse() != null) {
             var rWarehouse = runContext.render(this.getWarehouse()).as(String.class).orElseThrow();
-            properties.put("warehouse", rWarehouse);
+            properties.put("warehouse", quoteIdentifierIfNeeded(rWarehouse));
         }
 
         if (this.getDatabase() != null) {
             var rDatabase = runContext.render(this.getDatabase()).as(String.class).orElseThrow();
-            properties.put("db", rDatabase);
+            properties.put("db", quoteIdentifierIfNeeded(rDatabase));
         }
 
         if (this.getSchema() != null) {
             var rSchema = runContext.render(this.getSchema()).as(String.class).orElseThrow();
-            properties.put("schema", rSchema);
+            properties.put("schema", quoteIdentifierIfNeeded(rSchema));
         }
 
         if (this.getRole() != null) {
             var rRole = runContext.render(this.getRole()).as(String.class).orElseThrow();
-            properties.put("role", rRole);
+            properties.put("role", quoteIdentifierIfNeeded(rRole));
         }
 
         if (this.getPrivateKey() != null) {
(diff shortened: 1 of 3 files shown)


plugin-jdbc-snowflake/src/main/java/io/kestra/plugin/jdbc/snowflake/SnowflakeInterface.java
```diff
@@ -159,6 +166,32 @@ default void renderProperties(RunContext runContext, Properties properties) thro
...
+     * Ordinary names such as {@code INGESTION} are returned unchanged, so their existing
+     * case-normalization behavior is preserved. A value the user already quoted (starting and
+     * ending with a double quote) is left untouched. Any other value (for example one containing
+     * a colon like {@code INGESTION:TOPIC_ENVIRONMENT}) is wrapped in double quotes, with embedded
+     * double quotes escaped as {@code ""}.
+     */
+    static String quoteIdentifierIfNeeded(String value) {
+        if (value == null) {
+            return null;
+        }
```
jymaire, line 182: A blank rendered value now becomes `""` instead of being ignored.

`runContext.render(prop).as(String.class)` returns `Optional.of("")` when the value renders empty — the `Optional` is only empty when the `Property` itself is null. So `schema: "{{ vars.schema }}"` with an empty var used to put `schema=` in the connection properties, which the driver/server ignores. It now falls through to the quoting branch and puts `schema=""`, a zero-length quoted identifier, and login fails with an object-does-not-exist error.

Suggest short-circuiting on blank:

```java
if (value == null || value.isBlank()) {
    return value;
}
```

(or skipping the property entirely in `renderProperties` when the rendered value is blank).

jymaire (approved):

plugin-jdbc-snowflake/src/main/java/io/kestra/plugin/jdbc/snowflake/SnowflakeInterface.java
tchiotludo: Fixed — short-circuited on blank/null in `quoteIdentifierIfNeeded` so a blank rendered value is left as-is instead of becoming `""`. Pushed in 8afbc1f, added a unit test covering the blank case.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #1005, opened 2026-09-05
fix(db2): materialize BLOB/CLOB/NCLOB/XML instead of returning live locators

### LINKED ISSUE #1004: DB2 plugin: BLOB/CLOB/NCLOB/XML columns return live JDBC locators instead of materialized content
**Describe the bug**

`Db2CellConverter.convertCell()` returns the raw `java.sql.Blob`/`Clob`/`NClob`/`SQLXML` objects for these column types, instead of reading their actual content:

```java
case "blob" -> rs.getBlob(columnIndex);
case "clob" -> rs.getClob(columnIndex);
case "nclob" -> rs.getNClob(columnIndex);
case "xml" -> rs.getSQLXML(columnIndex);
```

These are live locators backed by the `ResultSet`'s underlying connection/statement, not self-contained values. Once rows are batched up for downstream processing (the normal pattern for a workflow engine reading query results) and the statement/connection has advanced or closed, calling `getBinaryStream()`/`getCharacterStream()`/`getString()` on that locator throws or returns garbage, depending on the DB2 JDBC driver's cursor-holdability configuration.

`OracleCellConverter` in the same plugin family already handles this correctly for BLOB/CLOB — it reads the full stream into a `byte[]`/`String` eagerly, while the `ResultSet` is still positioned on the row. The DB2 converter should do the same for BLOB/CLOB/NCLOB/SQLXML.

**Expected behavior**

BLOB columns should be returned as `byte[]`, and CLOB/NCLOB/XML columns as `String`, with content read immediately rather than deferred via a live locator.

**Additional context**

I have a fix ready (mirrors `OracleCellConverter`'s existing pattern, adds `Db2CellConverterTest` with mocked JDBC objects since the module's existing tests are all Testcontainers-based) and will open a PR referencing this issue.

karthikchundi-commits:
Fixes #1004

`Db2CellConverter.convertCell()` returned the raw `java.sql.Blob`/`Clob`/`NClob`/`SQLXML` objects directly for these column types — live locators backed by the `ResultSet`'s connection/statement. Reading them after the row has been batched for downstream processing, once the underlying statement/connection has advanced or closed, throws or misbehaves depending on the DB2 JDBC driver's cursor-holdability settings.

`OracleCellConverter` in the same plugin family already materializes BLOB/CLOB content eagerly for exactly this reason. Applied the same pattern here: read the full content into a `byte[]`/`String` while the `ResultSet` is still positioned on the row, then `free()` the locator.

Added `Db2CellConverterTest` with mocked `ResultSet`/`Blob`/`Clob`/`NClob`/`SQLXML` — this module's existing tests are all Testcontainers-based live-DB2 integration tests, so this is the first unit test for the converter itself, and runs without a live DB2 instance.

**Verified locally**: `./gradlew :plugin-jdbc-db2:test` — 5/5 pass (JDK 25, required by the current `io.kestra:core` dependency).

AI-assisted: implemented by Claude Code under my direction, following the existing `OracleCellConverter` pattern already established in this plugin family.

### DIFF
--- a/plugin-jdbc-db2/src/main/java/io/kestra/plugin/jdbc/db2/Db2CellConverter.java
+++ b/plugin-jdbc-db2/src/main/java/io/kestra/plugin/jdbc/db2/Db2CellConverter.java
@@ -2,14 +2,24 @@ package io.kestra.plugin.jdbc.db2;
 
 import io.kestra.plugin.jdbc.AbstractCellConverter;
 
+import java.io.ByteArrayOutputStream;
+import java.io.InputStream;
+import java.io.Reader;
+import java.io.StringWriter;
+import java.sql.Blob;
+import java.sql.Clob;
 import java.sql.Connection;
+import java.sql.NClob;
 import java.sql.ResultSet;
 import java.sql.SQLException;
+import java.sql.SQLXML;
 import java.time.ZoneId;
 import java.util.Calendar;
 import java.util.TimeZone;
 
 public class Db2CellConverter extends AbstractCellConverter {
+    private static final int LOB_BUFFER_SIZE = 8192;
+
     public Db2CellConverter(ZoneId zoneId) {
         super(zoneId);
     }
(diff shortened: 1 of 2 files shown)


Malaydewangan09 (changes requested):
Good catch, and the fix is the right shape. I checked `AbstractJdbcQuery`: FETCH/FETCH_ONE return rows after the connection is closed in the finally, and STORE cannot serialize a locator through Jackson either, so this was broken in all three fetch modes.

Two asks before merge.

**1. The `free()` in the `finally` can throw away the data you just read.**

If `blob.free()` throws, the finally replaces the return value and the task fails even though the read worked. And if both fail, the read error gets wrapped in `SQLException("Error reading BLOB data")` and then completely replaced by the free() exception, no cause, nothing suppressed. That is the broken-connection case this PR is about, which is exactly when you want the real error. Same in all four helpers:

```java
} finally {
    try { blob.free(); } catch (SQLException ignored) {
        // content is already read, a failed free should not fail the task
    }
}
```

**2. Can the four helpers move to `AbstractCellConverter`?**

Nothing in them is DB2 specific. `OracleCellConverter` has a near-copy inline, and Hana/As400 have the same bug. On the base class the siblings get it for free instead of copying 80 lines each.

Minor, not blocking: `getObject()` at the top of `convertCell` already returns the Blob, so `rs.getBlob()` fetches it a second time. That was free before this PR since nothing read the second one, but now both get materialized.

karthikchundi-commits:
@Malaydewangan09 Both addressed:

1. `free()` in each helper's `finally` is now wrapped in its own try/catch that swallows the `SQLException` -- content already read (or the real read error already captured) is no longer clobbered by a failed `free()`.
2. Moved `readBlob`/`readClob`/`readNClob`/`readSqlXml` onto `AbstractCellConverter` so `Db2CellConverter` calls up instead of owning private copies -- ready for Hana/As400 (#1014) to do the same rather than adding a third/fourth copy.

Added direct unit tests on the new base-class helpers in `AbstractCellConverterTest`, including the regression case (a mocked `free()` that throws must not discard already-read content, and a real read failure must still surface even when `free()` also fails). `plugin-jdbc`/`plugin-jdbc-db2` unit tests pass (58/58 + 6/6 non-Testcontainers).

Ready for another look.

Malaydewangan09:
@karthikchundi-commits the CI seems to be failing, can you check please?

karthikchundi-commits:
@Malaydewangan09 Checked — that failure was in `plugin-jdbc-clickhouse` (`Invalid LZ4 magic byte`), completely unrelated to this PR's `db2` module. `main` itself broke the same way that morning; your team fixed it directly on `main` a few hours later (pinning the ClickHouse image, then migrating Testcontainers to 2.x). Rebased this branch past that fix and pushed yesterday, but the triggered run is sitting in `action_required` — looks like it needs a maintainer to approve it before it'll actually execute. Let me know if you're able to approve it, or happy to push again if there's a better way to kick it.

Malaydewangan09 (approved):

PR #1014, opened 2026-09-15
fix(hana,as400): materialize BLOB/CLOB/NCLOB/SQLXML instead of returning live locators

karthikchundi-commits:

`HanaCellConverter` and `As400CellConverter` both had the same bug already fixed in `Db2CellConverter` (#1005): `blob`/`clob`/`nclob`/`xml` columns returned live `java.sql.Blob`/`Clob`/`NClob`/`SQLXML` locators straight from the `ResultSet`, instead of reading their content. A caller that reads the value after the row has moved on — e.g. once rows are batched for downstream processing, as this plugin's `Query`/`Queries` tasks do — hits an exception once the underlying statement/connection has advanced or been closed.

`As400CellConverter`'s own doc comment says it was *"Copied from the DB2 code as we cannot test AS400 we assume it works like DB2"* — that assumption is now stale, since DB2 no longer works this way after #1005.

This mirrors `Db2CellConverter`'s fix exactly: eager materialization via `getBinaryStream()`/`getCharacterStream()`, then `.free()` once read. `HanaCellConverter` only has `blob`/`clob`/`nclob` (no `xml` case), so it gets those three helpers; `As400CellConverter` has all four, so it gets `readSqlXml` too.

Found while auditing sibling converters in this plugin family for the same pattern after #1005 landed on Db2 — not filed as a separate issue since it's the identical, already-diagnosed bug in two more modules.

---

No live HANA or AS400/DB2-i instance available in this environment to run an end-to-end flow — same disclosed limitation as the original Db2/Oracle work in this repo. Verification is via `HanaCellConverterTest`/`As400CellConverterTest`, mirroring `Db2CellConverterTest`'s exact coverage shape (mocked `ResultSet`, asserting the returned value is the materialized content — not the live locator — and that `.free()` is called exactly once; a null-value case confirming `getBlob()` is never called once `getObject()` already reported no data).

```
./gradlew :plugin-jdbc-hana:test :plugin-jdbc-as400:test
```

9/9 new tests pass (JDK 25). The two pre-existing failures in `plugin-jdbc-hana` (`QueriesTest`/`QueryTest`) are a Testcontainers/Docker environment limitation in this sandbox, unrelated to this change — same limitation disclosed on the prior Db2/Oracle PRs.

---

### Contributor Checklist ✅


Malaydewangan09 (changes requested):
Same review as #1005, the fix itself is right. Two asks, both carried over from there.

**1. The `free()` in the `finally` can throw away the data you just read.**

If `free()` throws, the finally replaces the return value and the task fails even though the read succeeded. If the read fails too, its `SQLException("Error reading BLOB data")` gets completely replaced by the free() exception, no cause, nothing suppressed, which is the broken-connection case this PR exists for. Applies to all seven helpers here:

```java
} finally {
    try { blob.free(); } catch (SQLException ignored) {
        // content is already read, a failed free should not fail the task
    }
}
```

**2. This is now the third byte-identical copy of these helpers**, fourth counting Oracle inline. Worth moving them to `AbstractCellConverter` in #1005 and having all three converters call up. This PR only exists because the Db2 fix was not shared, so the same thing happens again with the next converter.

Also, merge #1005 first, the comment in `As400CellConverter` references it.

Minor, not blocking: `getObject()` at the top of both converters already returns the Blob/Clob, so `getBlob()` fetches it a second time. Cheap before, not cheap now that both get materialized. If you do use `data` directly, keep an `instanceof` guard, HANA can return `byte[]` from `getObject()` for small LOBs.

karthikchundi-commits:
@Malaydewangan09 Both addressed, same as #1005:

1. `free()` failures no longer clobber already-read content or mask a real read error -- fixed at the source now that it's shared (see 2).
2. Rebased onto the updated #1005 branch and dropped Hana/As400's own copies of `readBlob`/`readClob`/`readNClob`/`readSqlXml` entirely -- both converters now call the versions on `AbstractCellConverter`. No third/fourth copy.

All `HanaCellConverterTest` (4/4) and `As400CellConverterTest` (6/6) unit tests pass; the base-class helpers also have direct regression coverage in `AbstractCellConverterTest` now (free()-throws-but-data-survives, for all four LOB types).

Ready whenever #1005 is -- this branch is rebased on top of it, same order you flagged.

Malaydewangan09:
@karthikchundi-commits the CI seems to be failing, can you check please?

karthikchundi-commits:
@Malaydewangan09 Same root cause as #1005: unrelated `plugin-jdbc-clickhouse` Testcontainers breakage, already fixed on `main` by your team, rebased this branch past it and pushed yesterday. Details in my reply there: https://github.com/kestra-io/plugin-jdbc/pull/1005#issuecomment-5781443097 — this run is also stuck in `action_required` pending approval.

Malaydewangan09 (approved):

